PR #2, opened 2021-08-27
clippy: ptr_as_ptr

wcampbell0x2a:
## Motivation

Change casts that don't change mut to `.cast<T>()`.


bastibl:
Great, thank you very much! I really need to look into clippy and extend CI to run these checks.

PR #9, opened 2021-08-30
feature/ZeroMQ

### LINKED ISSUE #3: ZeroMQ PubSink & SubSource
**Why ZeroMQ?**
ZeroMQ is by default in GNU Radio Companion.

**ZMQSubSource & ZMQPubSink as blocks**
I would like to have blocks to publish stream & receive stream from other systems.

**Underlying ZMQ rust lib?**
I have done [an experiment](https://github.com/loic-fejoz/FutureSDR/tree/feature/ZeroMQ) with [zmq](https://crates.io/crates/zmq) library which is blocking one.
Should we consider using [async-zmq](https://crates.io/crates/async-zmq) (non blocking) or the more idiomatic [libzmq](https://crates.io/crates/libzmq) (but unmaintained).

**Additional context**
Having those would enable tests where only part of a chain is migrated to FutureSDR. For instance, it could enable to split a chain between GNU Radio and FutureSDR until FutureSDR  had all the mandatory blocks of the overall chain.

loic-fejoz:
Add ZMQSubSource and SMQPubSink for streaming over ZeroMQ protocol which is also by default in GNU Radio.
I think I have made it as a feature activated by default.
It also come with two binaries (one receiver, one emitter) that can talk to each other and/or with GNU Radio.

Apparently the topics is not settable in GNU Radio thus I have not provided way to set it. Maybe I should?

Close #3

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ members = [
     "examples/android",
     "examples/logging",
     "examples/wasm",
+    "examples/zmq",
     "perf/msg",
     "perf/null_rand",
 ]
@@ -37,6 +38,7 @@ anyhow = "1.0.38"
 async-trait = "0.1.41"
 clap = "2.33.1"
 config = "0.11.0"
+cpal="0.13.4"
 dirs = "3.0.1"
 futures = "0.3.5"
 futures-lite = "1.10.0"
@@ -49,6 +51,7 @@ rand = "0.8.0"
 slab = "0.4.4"
 spin = "0.9.0"
 serde = { version = "1.0", features = ["derive"] }
+zmq = "0.9"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 wasm-rs-async-executor = "0.9.0"
@@ -91,6 +94,7 @@ winapi = { version = "0.3", features = ["sysinfoapi", "winbase", "handleapi", "m
 async-channel = "1.6.1"
 async-executor = "1.3.0"
 easy-parallel = "3.1.0"
+#clang-sys = "1.2.1"
 
 [profile.release]
 codegen-units = 1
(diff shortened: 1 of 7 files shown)


bastibl:
Awesome! With example and even feature-gated. Really cool :-)

I'll merge this as is and do some minor follow-up patches (mainly because I'll be afk the next days and want to avoid stalling this through a lot of back-and-forth).

Regarding error handling: in the functions that return a `Result<X>`, you can use the `?` operator to deal with errors. In case of an `Option<X>`, you could use `.context("option was none")?` instead of `expect()` or `assert!(....is_ok())`. 

There are also some related questions, I thought about. Maybe you have some thoughts here as well:
- I'm not sure if such blocks should be typed, i.e., `ZmqSubSource(4, ...)` or `ZmqSubSource<u32>(...)`. I think at some point, we might have to switch to typed blocks, since the alignment of data in the buffer cannot only be determined by the size of the data (see [align_of](https://doc.rust-lang.org/std/mem/fn.align_of.html)). I'll also have adapt `Copy` and `NullSource` etc.
- I also prefer `ZMQSubSource` but it seems to be idiomatic Rust to write `ZmqSubSource`. It's also why I called PMTs `Pmt`s. But in this case, we could anyway get around this by putting the blocks in a submodule/subfolder and have `zmq::SubSource`. This also eases feature-gating and (IIRC) there are other transfer modes, so we might add more blocks in the future.
- I'm not sure what to do with default features. Since compile times tend to get large, I thought about keeping them minimal and also remove soapy and vulkan. The binaries and examples can be annotated with `required-features` so one will get a nice message, telling the user to enable the feature. Unfortunately, there doesn't seem to be a user config to overwrite default features, for example, if you work on the ZMQ stuff. So the only option would be to edit `Cargo.toml` temporary and add the feature.

Thanks again!

loic-fejoz:
Concerning having typed blocks, I have mix feeling. Somehow I think the runtime could do the transformation and thus we can have "agnostic" blocks. But on the other side, It would probably increase overhead and thus may not be a good idea.
So maybe we should have transformation blocks moreover with typed blocks indeed.

NB: Thanks for the tip about `context` (I am a rust newbie).

PR #10, opened 2021-08-31
feature/constblock

loic-fejoz:
## Motivation
Because GNU Radio tutorials are using basic blocks applying some constant operation (AddConst, MulConst), we should add this feature. See also the [discussion 6](https://github.com/FutureSDR/FutureSDR/discussions/6).

## Solution

* Shared code for operations requiring 1 constant.
* Implementation for Addition and Multiplication based on closure.

### DIFF
--- a/src/blocks/add_const.rs
+++ b/src/blocks/add_const.rs
@@ -1,0 +1,77 @@
+use anyhow::Result;
+use std::cmp;
+use std::ops::Add;
+
+use crate::runtime::AsyncKernel;
+use crate::runtime::Block;
+use crate::runtime::BlockMeta;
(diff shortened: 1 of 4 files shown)


bastibl:
What a great idea to make it generic over a function! Why didn't I think of that :-) I adapted it slightly so that it is a block with arbitrary `FnMut(A) -> B`. I hope that's fine for you. All changes are on top of your commits. (GitHub doesn't seem to recognize that these commits are in-tree, probably because I rebased them. Sorry -.-)

bastibl:
I called the adapted block `Apply`, see here: https://github.com/FutureSDR/FutureSDR/blob/master/src/blocks/apply.rs#L14-L20

PR #11, opened 2021-08-31
Feature/soapysdr args

wcampbell0x2a:
## Motivation

SoapySDR inputs are hardcoded and should be user defined.

### Add custom freq/sample_rate/gain SoapySdr Block
Add freq/sample_rate/gain to the SoapySdrBuilder. This allows the freq/gain/sample_rate to be set in the init() function once the soapySDR library has the context of the device.

This was tested on a HackRF One device.

### Enable soapysdr logging

This  changes the following output:
```
[INFO] Opening HackRF One #0 57b068dc214b4c63...
```

into the following:
```
FutureSDR: INFO - Opening HackRF One #0 57b068dc214b4c63...
```

### DIFF
--- a/examples/soapy.rs
+++ b/examples/soapy.rs
@@ -41,7 +41,11 @@ mod inner {
     pub(super) fn main() -> Result<()> {
         let mut fg = Flowgraph::new();
 
-        let src = SoapySourceBuilder::new().build();
+        let src = SoapySourceBuilder::new()
+            .freq(100e6)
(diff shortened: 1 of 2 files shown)


bastibl:
Nice, thank you!
(I messed with master, introduced a conflict, fixed, and pushed manually. Somehow it doesn't recognize, so I'll have to close the PR. But your commits are in-tree. Sorry for the hassle -.-)

Issue #4, opened 2021-08-29
AudioSource & AudioSink


loic-fejoz:
**AudioSource & AudioSink**
The AudioSource & AudioSink blocks are one of the basic blocks of a lot of tutorial.

**underlying library?**
[cpal](https://crates.io/crates/cpal) could be a candidate as the underlying library.
Being a newbie in Rust, I do not know much about the state of such audio crates.
NB: I have tried to provide such a block but I am stuck with some requirement from cpal of [having 'static callbacks](https://github.com/loic-fejoz/FutureSDR/blob/feature/audio-source/src/blocks/audio_source.rs#L41).

bastibl:
I just pushed some audio blocks (source, sink, oscillator, file source), in case you want to give them a try.

PR #13, opened 2021-09-14
FM Receiver example

loic-fejoz:
## Motivation

See how FutureSDR can be applied on simple example such as an FM Receiver.
Also testing the AudioSink as stated in #4 

## Solution

While we add discussion on Discord and other alternatives implementation, we come up with the idea that an Apply block with a mutable closure would be sufficient.


bastibl:
Hi, I hope you don't mind. But since I didn't want this to be stalled, I just added a very minimal FM receiver example to the project: https://github.com/FutureSDR/FutureSDR/blob/master/examples/fm-receiver/src/main.rs

PR #14, opened 2021-10-16
Added basing application doing WBFM demodulation using liquid-dsp

mryndzionek:
An example performing WBFM demodulation.
The example app has to blocks:
 - [Resampler](https://github.com/mryndzionek/FutureSDR/blob/liquid_integ/examples/sdr/wbfm.rs#L76)
 - [WBFMDemod](https://github.com/mryndzionek/FutureSDR/blob/liquid_integ/examples/sdr/wbfm.rs#L205)
 
Let me know if this approach (using liquid-dsp via bindgen) is acceptable. The example is fully functional - I can listen to FM radio stations using my RTLSDR. There are audio buffer underruns, but probably due to audio sink misconfiguration.


mryndzionek:
I agree that FFI bindings should be separated. I don't know if you seen, but the [crate](https://github.com/jholtom/liquid-dsp-bindings) you mention is essentially empty. It only pulls in the bindgen bindings from [here](https://github.com/jholtom/liquid-dsp-bindings-sys). Both crates don't seem to be actively maintained either.

I can create a separate repo with my bindgen bindings, but not sure about the high-level interface. I'm still learning Rust and I'm not sure how much time I will be able to spend on this in the future. Maybe you could add a repo for this in [FutureSDR](https://github.com/FutureSDR) with maybe a sketch of a interface you envision?

Maybe we could start adding blocks using FFI bindings directly and refactor as we go?

mryndzionek:
Small update. I've found two more repos that seem to be more complete regarding the high-level interface:
 - https://github.com/cubehub/rust-liquid-dsp - [interface example](https://github.com/cubehub/rust-liquid-dsp/blob/master/src/msresamp.rs#L35)
 - https://github.com/RamiHg/rust-liquid-dsp - [interface example](https://github.com/RamiHg/rust-liquid-dsp/blob/master/src/filter.rs#L28)
 
Let me know what you think.

PR #26, opened 2021-12-22
FIRFilter Implementation as well as Decimation and simple Quadrature Demodulation

justacec:
I have pushed here an early version (read as non-optimized) version of FIR filter along with simple Decimation and Quadrature Demodulation blocks.  My intent was to generate the necessary code to begin processing the RDS data stream in the FM broadcasts.  The submitted code did function to generate a clean signal with the following flow graph:

```
let src = FileSourceBuilder::new(mem::size_of::<Complex<f32>>(), "test.dat".to_string()).build();
let hd = HeadBuilder::new(mem::size_of::<Complex<f32>>(), 10000*4096).build();
let decimation = DecimationBuilder::<Complex<f32>>::new()
    .sample_rate(2.56e6)
    .decimation(12)
    .build();
let fir = FIRFilterBuilder::new()
    .fft_width(2048)
    .response_type(FIRFilterResponseShape::LowPass(100000_f32))
    .impulse_size(300)
    .sample_rate(2.56e6/12_f32)
    .build();
let demod = QuadratureDemodulation::new();

let src = fg.add_block(src);
let hd = fg.add_block(hd);
let decimation = fg.add_block(decimation);
let fir = fg.add_block(fir);
let vsnk = fg.add_block(vsnk);
let demod = fg.add_block(demod);

fg.connect_stream(src, "out", hd, "in")?;
fg.connect_stream(hd, "out", decimation, "in")?;
fg.connect_stream(decimation, "out", fir, "in")?;
fg.connect_stream(fir, "out", demod, "in")?;
fg.connect_stream(demod, "out", vsnk, "in")?;
```

I am not suggesting that this code is ready for merge, but could be a good discussion starter.  I realize that there has been a recent addition of the FIR filter.  Maybe my attached code could help to make it more user friendly / flexible.  :)


bastibl:
Sorry, I'm just catching up after the Christmas break. My current idea would be to have this as an alternate FIR filter implementation. To other implementation requires nightly features, which I don't want to force for the whole project. So there needs to be another implementation that is selected depending on the toolchain.

Did you, by chance, continue with the receiver? It would be nice to have an example for the new blocks.

Btw. for the quadrature demodulator, you could also use the Apply block and do something like (untested):
```rust
let mut last = Complex32::new(1.0, 0.0);
let demod = Apply::new(
    move |&c: Complex32| -> f32 {
        let arg = (c * last.conj()).arg();
        last = c;
        arg
   });
```
There are some minor issues with the code, but I don't want to annoy you with a lot of back-and-forth. I'd suggest to merge this (and maybe the updated example) and do some follow-up work, trying to combine with the other implementation. Would this be fine for you?

bastibl:
Hi, I just added a minimal FM receiver example that demonstrates the idea about implementing the demodulator with the Apply block: https://github.com/FutureSDR/FutureSDR/blob/master/examples/fm-receiver/src/main.rs#L23-L28

Also, we just reworked the FIR filter API and plan factor out the DSP parts in a separate library. In this library we should add the tap generators that you have in your PR. I don't have a concrete idea, but maybe something like `FirTaps::low_pass<T: TapType>(x, y, z) -> Vec<T>` or something similar to the enums that you defined in your code.

I'm not sure how to proceed with the PR. Are you still interested?

justacec:
HA!  Funny I was just thinking about this yesterday.  Was planning on
pulling the thread on it this weekend.  I am still interested and would
love to be a part.  I have a LimeSDR coming in the Mar delivery.  Let me
look more at this over the weekend.  Thanks for following up.

justacec:
Do you think the Taps need to be broken out into a separate library?  I guess I would be interested in your overall vision.  I like the idea of a single library that covers everything related to SDR.  But, in the end I might guess that would get too cumbersome.

bastibl:
We want to try, putting DSP stuff in a separate crate that we keep (for now) in this repo. The first step is taken here: https://github.com/FutureSDR/FutureSDR/pull/33
Ideally, it would become something like a Rust version of liquiddsp.

lkolbly:
Having a `firdes` submodule of `futuredsp` I think would work really well, and would be similar to what `liquiddsp` does. (Currently my FM radio is 508 lines, of which 318 are hardcoded taps copied from fiiir.com)

bastibl:
I'll close this for now, as there were quite a few changes since this PR. We would, of course, appreciate, if you had the time to follow-up on this work at some point.

PR #28, opened 2022-01-30
Support multi-byte items in FileSink and FileSource

lkolbly:
This change adds a `T` parameter to `FileSinkBuilder` and `FileSourceBuilder` to set the item size for the built objects. Additionally, this PR modifies the blocks to support item sizes other than 1. For the most part, the only change needed to support that was properly dividing by the item size when advancing the streams - although the blocks access the streams as slices of `u8`, the streams are internally aware of the item size.

Additionally, for `FileSource`, this PR makes the source silently finish when it received EOF, rather than panic. `read_exact` doesn't tell us how many bytes are read, so we are forced to drop whatever bytes were read and finish the stream. (conceivably, we could rewrite the algorithm to use `read` directly, but one step at a time)

### DIFF
--- a/src/blocks/file_sink.rs
+++ b/src/blocks/file_sink.rs
@@ -14,7 +14,7 @@ use crate::runtime::StreamIoBuilder;
 use crate::runtime::WorkIo;
 
 pub struct FileSink {
-    // item_size: usize,
+    item_size: usize,
     file_name: String,
     file: Option<File>,
     n_written: usize,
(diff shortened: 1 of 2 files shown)


bastibl:
Looks like I missed quite some messages during the weekend :-) Very cool that you solved the problem. I completely forgot about this limitation. I merged this and also pushed follow-up commits to make more blocks generic over the data type (Null source/sink, head, throttle, etc). I also gave it a try with the `read` call as you suggested. It's not heavily tested, but it seems to work. (Also I deleted some builders that didn't make any sense, since the block only took a fixed set of parameters, so you might have to adapt scripts slightly.) Let me know if you encounter any issues and thanks for your contribution!

PR #29, opened 2022-02-04
Generic FIR filters

lkolbly:
This PR changes the FIR API a bit, so that it's unified between `fir_stable` and `fir`. It also adds support for multiple sample types (adding `Complex<f32>`), and in the future could potentially be extended to support other types. And there's support for both dynamic and static taps. (with `[f32; N]` taps, I didn't see any slowdown over the previous implementation when compiled with `+nightly`)

In particular, the actual FIR kernel portion is factored into a separate trait, and the FIR block itself is simply a templated wrapper around it which handles the IO portion. (really, at a high level, the `Fir` struct is now just an optimized version of the `Apply` struct. But for now I haven't thought that far) Right now, all `Fir` blocks use the same `FirKernelCore`, however I'm imagining that in the future there could be SIMD implementations and resampling implementations and so on and so forth. (I'd be willing to maintain `FirKernel` as a separate crate, if you don't think it belongs here. Conceivably it could eventually get quite far into the weeds)

Anyway. There's a couple things in particular I'm not the hugest fan of, mostly the code duplication all over the place, but I thought I'd at least post this and see if it's in the right direction.


lkolbly:
Ah, the CI's telling me I missed updating the `perf/` code. I can look into that this evening.

bastibl:
Wow, this is awesome. And even with documentation :-) Thank you!

To be honest, I didn't even know that most of this is possible. I am still not sure, if I understand the details of the current type system -.- So there's definitely some additional complexity. But the new user-facing interface is much better so that's probably worth it. The only thing I wondered is, if it would be possible to hide the third type parameter for the builder somehow (like `impl TapsAccessor`). I toyed a bit around, but didn't find a solution.

One bug is the termination logic. It should terminate, if the input is finished and it consumed all possible samples (`i.len() - num_taps + 1`). For this, the `work()` function has to know the taps, so it probably should be added to the `FirKernel` interface). I'll push my suggestion to this PR.

I ran the FIR performance evaluation again with the changes and also didn't note any differences. So I guess a nightly build still uses const generics and the fast add/mul operation. That's also my interpretation of the gdb disassembler output.

Regarding the question where this code should live: I think it would be awesome to have a Rust DSP library for SDR, similar to liquiddsp but with all the cool stuff and ergonomics that are possible with Rust. I didn't find something like that. Most existing Rust DSP libraries cannot be used for SDR, since they either allocate `filter(...) -> Vec<_>`, i.e., do not allow to specify an output buffer; or are based around iterators `filter(...) -> impl Iter<Item=_>`. Both approaches don't go together with our use case.
Creating such a library would be great and helpful way beyond FutureSDR. If you are interested, we could start the library as a sub-crate, living in this repo's workspace, similar to the frontend and pmt. Or we could start a separate crate under the FutureSDR org. The former would ease keeping everything in sync. In the beginning there's probably a lot in the flow.

Let me know, if you're fine with my fix. Then I'd go ahead and merge.

lkolbly:
> One bug is the termination logic. It should terminate, if the input is finished and it consumed all possible samples (i.len() - num_taps + 1). For this, the work() function has to know the taps, so it probably should be added to the FirKernel interface). I'll push my suggestion to this PR.

Looks generally good to me, although I did push a `num_taps` implementation for the stable implementation of `FirKernel<Complex<f32>>`.

One question is, can we tell if that condition is met, based on `consumed` and `produced`? I'm thinking, if the input stream is finished, and `produced < o.len()`, then it should be the case that the fir kernel has consumed everything it ever will, because it has space to produce more samples and it's not going to receive any more input samples.

> The only thing I wondered is, if it would be possible to hide the third type parameter for the builder somehow (like impl TapsAccessor). I toyed a bit around, but didn't find a solution.

Well, so, the `TapsAccessor` knows what type the tap is, which means I think we could do an API where you specify just the `SampleType` and `Taps`:
```rust
let fir = FirBuilder::new::<f32, [f32; 3]>([1.0, 2.0, 3.0]);
let fir = FirBuilder::new::<Complex<f32>, &[f32; 3]>(&[1.0, 2.0, 3.0]);
let fir = FirBuilder::new::<f32, Vec<f32>>(vec![1.0, 2.0, 3.0]);
```
or, of course, just:
```
let fir = FirBuilder::new::<f32, _>(vec![1.0, 2.0, 3.0]);
```

> we could start the library as a sub-crate, living in this repo's workspace, similar to the frontend and pmt

I think this is the best idea, because yeah it's definitely going to be useful to be able to keep these in sync. I did look around and found much the same things as you did - certainly an iterator-based approach has better latency, but I don't think that's really what we want.

bastibl:
> One question is, can we tell if that condition is met, based on consumed and produced? I'm thinking, if the input stream is finished, and produced < o.len(), then it should be the case that the fir kernel has consumed everything it ever will, because it has space to produce more samples and it's not going to receive any more input samples.

Yes, if `produced < o.len()` it should have consumed all input samples. The problem is if `produced == o.len()` it could *also* have consumed all input samples. I think, it can only be differentiated knowing the taps. For more general kernels (you mentioned resamplers), it might become more difficult. But I'd say, we figure this out later. Same for the other API experiments. Let's merge this.

Any ideas for a good name for the DSP library?

lkolbly:
> Any ideas for a good name for the DSP library?

`futuredsp`? Or `futuresdr-dsp`. Both in keeping with the "future" theme.

PR #30, opened 2022-02-04
Add default AudioSink constructor

phlorian:
This PR solves an issue I had with the AudioSink block. My audio device does not support 48 kHz sample rate. The AudioSink would simply block without an error message.

My approach is to provide a different new function that will use the default settings of the default device.

This PR is just a simple fix to make the examples work. Better error handling would probably also help.

I am looking forward for any feedback.


bastibl:
Thanks for the PR. How about, having `new()` with no arguments (what you have as `new_default`) and also implement the `Default` trait for it (I think clippy will also ask for this). The current constructor could be something like `with_config`. Or we add a builder instead.

As far as I see, the problem with the default constructor is that it needs more getters to check what parameters are in use and configure other blocks accordingly. But these blocks are anyway very limited at the moment. So we could do this in a follow-up commit.

phlorian:
Thank you for your comment.

I forgot to mention that I am new to Rust. So I might miss obvious things like the `Default` trait.

Changing the code to `new` and `with_config` is easy. This is what I have done and pushed.

Is implementing the `Default` trait for `AudioSink` the right thing? This will return just a default `AudioSink` not a `Block`. the `new` and the `with_config` functions are returning a `Block` crated by `Block::new_async`.

If you prefer that I go for a `AudioSinkBuilder`, then I will try to do this.

bastibl:
Arg, sorry. Please forget about the `Default` trait that was nonsense.

But I think the other issue remains: The current code sets the audio rate explicitly, e.g., it makes sure that the sample rate of the audio sink matches the sample rate of the mp3. With this commit, we might get anything and rates might not match. A full example would have to check both rates (i.e., the rate of source and sink) and add resamplers as needed. I think, until we have the full-fledged example, we have to stick to the explicit setting.

phlorian:
Yes, I agree that this is not a good thing with the sample rate.

As this PR was meant to be a quick fix that it plays at least a tone, I suggest to close this PR. I will create a new one if I have a better solution. As you mentioned this should include a way to resample incoming data to the device sample rate.

Another thing is throttling. The test mp3 is playing at full speed. Funny but not correct. :-) Is the speed correct for you? Do you think that the AudioSink should throttle automatically or is this always in the responibility of the sender?

bastibl:
I think, until we extend it with resamplers and more logic, the best thing we can do is to detect the limitation of the audio device and error/panic with an appropriate message. Maybe you know a way to detect this?

Regarding throttling: this is the task of the audio device. If the test mp3 plays too fast, this an issue with the sample rate of the audio device. I assume, your audio card used a very high sample rate.

phlorian:
I removed the changes. This seems to be a cpal problem on Windows. This should not be "fixed" in FutureSDR. Sorry about that. For now it will panic if the configurations do not match.

bastibl:
Thanks, for PR! It looks like the runtime needs better error handling if this error was not propagated properly -.- But that's on my todo list, also introducing custom error types and not abusing anyhow::Result for everything :-)

PR #33, opened 2022-02-19
Create futuredsp crate

lkolbly:
This creates the `futuredsp` crate, which exists to contain high-throughput implementations of core DSP algorithms in a convenient-to-use API.

Currently it only contains a non-resampling FIR filter (real and complex samples with real taps). Additionally it has a separate benchmark setup to measure the algorithm's performance directly.


lkolbly:
Closing since this is a subset of #35

PR #34, opened 2022-02-21
Remove destructuring assignment in slab allocator

lkolbly:
```
fixup! refactor slab buffer

   `_ = ...` is a "destructuring assignment", which as of 1.58.1 is behind a
   feature flag. However, in nightly it is stabilized, and so this code works
   on nightly, but not on stable.

   This commit changes the `_ = ...` to `let _ = ...`, which I think has the
   same behaviour.
```


bastibl:
Thanks that was a typo. Didn't even know that it works without `let` :-)

PR #35, opened 2022-02-21
Change how FIR filters detect the end condition

lkolbly:
This PR is a follow-on to #33 (and this contains the same commits as that PR, because I couldn't figure out how to make a PR to merge this into that). But, it's conceptually different so I probably not worth combining the PRs.

So I think we can move the finish condition into the FIR filters by calculating how many more samples we _could_ have produced. If the input is finished, and the core could not have produced more samples (even if it had room), then I think we can finish the block?

If so, this'll make resampling filters work a bit easier, eventually.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ categories = ["asynchronous", "concurrency", "hardware-support", "science", "was
 [workspace]
 members = [
     ".",
+    "futuredsp",
     "pmt",
     "frontend",
     "examples/android",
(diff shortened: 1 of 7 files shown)


bastibl:
Thanks, I was also thinking about this issue recently. I think giving back concrete numbers might not always be straightforward and is also not really needed. How about returning an enum with `BlockedIn`, `BlockedOut`, and `Done` to indicate that the input/output was the limitation or if everything could be processed?
Since this includes the other PR, I'd wait for this discussion and merge this PR, if that's fine for you.

lkolbly:
> How about returning an enum with BlockedIn, BlockedOut, and Done to indicate that the input/output was the limitation or if everything could be processed?

Okay, yeah that makes sense. I'm not entirely sure though what the right enum names are - `Done` in particular doesn't always apply, since it could happen even if the block isn't finished. What about `InsufficientInput`, `InsufficientOutput`, and `BothSufficient`?

bastibl:
Thanks, I don't have better suggestions for the names. I'd say, we just go for it.

PR #36, opened 2022-02-24
Consider no_std in futuredsp

phlorian:
This PR is changing the fir in futuredsp to no_std.

The motivation is to keep this crate as platform independent as possible, especially also ready for bare metal programming. More infos about it here: https://docs.rust-embedded.org/book/intro/no-std.html

### DIFF
--- a/futuredsp/benches/benchmarks.rs
+++ b/futuredsp/benches/benchmarks.rs
@@ -3,6 +3,9 @@ use futuredsp::fir::{FirKernel, NonResamplingFirKernel};
 use num_complex::Complex;
 use rand::Rng;
 
+extern crate alloc;
+use alloc::vec::Vec;
+
 trait Generatable {
     fn generate() -> Self;
 }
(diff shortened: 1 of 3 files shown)


bastibl:
Good point. It would be nice to have and it should definitely be a long-term goal. Also the general approach to provide pointers to input/output buffers should align very well with `no_std`.
At the same time, I have no experience working with `no_std` code. This is just the very beginning of the attempt to factor out DSP in a separate crate. I assume well go back and forth with implementations. I wonder how much friction `no_std` introduces here. If it makes implementation and refactoring more painful, it's definitely not worth it. For the FIR, it seems to be rather straightforward, but for other algorithms, it might be more complicated. Also, the library is very immature and tight to FutureSDR, which requires `std`. So there is no immediate benefit from doing this now.
Do you have more experience with `no_std` code? Do you think it is a good idea to do this now? Do you think retrofitting it at a later stage is possible?

wcampbell0x2a:
`no_std` usually has the following, with default-features holding `["std"]`:

https://github.com/sharksforarms/deku/blob/b9a5f7b21b6bd597dd7602b8dcb8f2d2c7697d68/Cargo.toml#L21

One would use `default-features = false` and enable the `alloc` and/or `std` features.

bastibl:
Thank you all for your input. So let's merge this.

PR #37, opened 2022-02-27
Add transfer example

lkolbly:
The `SampleCounter` allows stopping the flowgraph after exactly N samples have passed. This is particularly useful for the `record` example which is also added, which simply grabs samples from a soapy source and stuffs them in a file. It also prints out average magnitudes and warns when the magnitude is close to 1.0. I'm not really sure if that's the best way to do that (since my understanding is the better way is to watch the noise floor vs. the signal), but in a generic IQ stream it's hard to do better. But I'm also fine pulling it out.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ members = [
     "examples/fm-receiver",
     "examples/logging",
     "examples/spectrum",
+    "examples/transfer",
     "examples/wasm",
     "examples/wgpu",
     "examples/zeromq",
--- a/examples/transfer/Cargo.toml
+++ b/examples/transfer/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "transfer"
+version = "0.1.0"
+edition = "2021"
+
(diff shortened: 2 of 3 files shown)


bastibl:
Hi, if I understand that correctly, that should be what the `Head` block is doing. (Sorry, we need more documentation.) The other functionality sounds like it could be an AGC block that (optionally?) adapts the gain as needed (by sending PMTs to the SDR Source).

lkolbly:
> that should be what the Head block is doing

Looking at it, yeah, I think so.

> we need more documentation

I need to read through the blocks that already exist :P I can write up some docs for that block though.

> The other functionality sounds like it could be an AGC block

Yeah, I guess it is the power-measurement part of an AGC block.

bastibl:
Thanks for the update. Let me know, when you think it is good to go. I'd be happy to merge as-is and make a follow-up to exclude it from mac/win CI tests and maybe rename it to `rx-to-file` or similar.

bastibl:
I just merged it with the mentioned changes as follow-up commits. GitHub doesn't seem to get it, but the commits are in `master`. Thank you very much!

PR #39, opened 2022-03-12
Add some docs!

lkolbly:
Inspired by the [nom docs](https://github.com/Geal/nom/blob/main/doc/choosing_a_combinator.md), I started writing some block docs. I've only written a few so far, but I think it's a reasonable cross-section of blocks (and I plan to write more later but I have to go get food right now).

The high points:
* There's a summary table in `futuredsp::blocks` that looks quite nice, and gives a brief description of all the available blocks, as well as sorting them by type.
  * Note that the major difference between this and the block listing below is twofold: One, it groups the blocks, and two, it excludes things that aren't blocks, like `MessageSourceBuilder` vs. `MessageSource`. But at the same time, it does duplicate quite a bit of information.
* Each block's docs has a standardized:
  * Description,
  * list of inputs & outputs (including marking which are message inputs/output),
  * and usage code sample.
* Also, the `doc(cfg())` gives us the nice little purple tag that says what features are required. There is a `doc_auto_cfg` feature, but it doesn't seem to pick up the features well, so in my limited experimenting we do have to manually specify it. The method for enabling it on docs.rs is stolen from [StackOverflow](https://stackoverflow.com/questions/61417452/how-to-get-a-feature-requirement-tag-in-the-documentation-generated-by-cargo-do).

To generate the docs locally, you can run e.g.:
```
$ RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc
```

Summary:
image

Block documentation:
image

(also, note that some of the links for the FIR documentation are currently broken, because they refer to some things I'm still working on in another branch)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -156,3 +156,6 @@ debug = true
 lto = "fat"
 opt-level = 3
 panic = "abort"
+
+[package.metadata.docs.rs]
+rustdoc-args = ["--cfg", "docsrs"]
--- a/src/blocks/apply.rs
+++ b/src/blocks/apply.rs
@@ -11,6 +11,43 @@ use crate::runtime::StreamIoBuilder;
 use crate::runtime::SyncKernel;
 use crate::runtime::WorkIo;
 
+/// Applies a function to each sample in the stream.
+///
+/// # Inputs
+///
+/// `in`: Input
+///
+/// # Outputs
+///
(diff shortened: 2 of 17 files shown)


bastibl:
Wow, this is so awesome! I can't wait to make a new release to get this out. Thank you so much.

PR #40, opened 2022-03-12
Accept Into<String> for SoapySourceBuilder filter

lkolbly:
Just a little change I found while going through the docs


bastibl:
Thanks!

PR #41, opened 2022-03-12
Round FileSource output to item_size

lkolbly:
This change enforces that, as long as there are sufficient bytes, a
multiple of item_size bytes will be read from the file.

Note that the block will always _produce_ a multiple of item_size bytes, but I don't think it was guaranteed that it would always read a multiple. (for example, if the `item_size` is 8, and the output buffer is 11 bytes long) Note that I haven't actually reproduced or encountered this bug, but I think it is actually a bug.

### DIFF
--- a/src/blocks/file_source.rs
+++ b/src/blocks/file_source.rs
@@ -46,6 +46,9 @@ impl<T: Send + 'static> AsyncKernel for FileSource<T> {
         let out = sio.output(0).slice::<u8>();
         let item_size = std::mem::size_of::<T>();
 
+        let out_len = out.len();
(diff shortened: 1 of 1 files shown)


bastibl:
I think it *should* be fine, since the stream port is actually of type `T`, but it is not obvious from the code. Ideally, this would somehow be reflected in the type system. But I didn't find a way to achieve this. At the moment, one can cast stream buffers to anything :-/ One could add an `assert!` to make sure this invariant actually holds. Or one could cast the buffer to type `T` and afterwards cast to byte. Do you have any preferences?

PR #42, opened 2022-03-13
Add IIR filter

lkolbly:
This refactors some of the `futuredsp` code to make the kernel interfaces generic, and adds a IIR filter block.

It isn't currently optimized (at all), and it also only supports `f32` samples with `f32` taps. But, on my mid-tier laptop, it does apparently run at >70MS/s with 7 feedback taps and 1 feedforward tap, which is pretty reasonable, and certainly good enough for e.g. a FM radio de-emphasis filter or other audio work.


bastibl:
Wow, nice! The abstraction with the different kernel types is cool. Thank you so much for all your contributions!

PR #43, opened 2022-03-14
Basic FIR design functionality

### LINKED ISSUE #32: logging should use the tracing library

AndersKaloer:
This implements basic window-based methods for designing FIR lowpass, highpass and bandpass filters.

The functionality is contained in the following modules:
- `futuredsp::windows`: Contains a generic `FilterWindow` trait and a number of commonly used window functions.
- `futuredsp::firdes`: Contains general lowpass, highpass, and bandpass filter design methods based on the window method.
- `futuredsp::firdes::kaiser`: Contains methods for generating lowpass, highpass, and bandpass filters by specifying cutoff frequencies, transition bandwidths, and maximum ripple. The design is based on the Kaiser method.
- `futuredsp::math::consts`: Replacement for `std::f32::consts` and `std::f64::consts`. So far only used for pi.
- `futuredsp::math::special_funs`: Special functions. So far includes an approximation of the modified Bessel function of the first kind of order zero. The approximation appears to be the same as the one used in GNU Radio.

An example is provided in `examples/firdes`.

Feel free to propose changes to this structure.

The implementations of the various functions have been tested against Matlab methods for a few test cases. I have left a test case for most of the functions in the code.

I am missing the possibility to print warnings in a couple of places. I have left a couple of commented `println!()` where it would be useful. I guess the best way to resolve these issues is to fix #32 and include it in `futuredsp`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ members = [
     "examples/android-hw",
     "examples/audio",
     "examples/cw",
+    "examples/firdes",
     "examples/fm-receiver",
     "examples/logging",
     "examples/rx-to-file",
--- a/examples/firdes/Cargo.toml
+++ b/examples/firdes/Cargo.toml
@@ -1,0 +1,8 @@
+[package]
+name = "firdes"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+futuresdr = { path = "../..", features = ["audio"] }
+futuredsp = { path = "../../futuredsp", version = "0.0.2" }
--- a/examples/firdes/src/main.rs
+++ b/examples/firdes/src/main.rs
@@ -1,0 +1,58 @@
+use futuresdr::blocks::audio::AudioSink;
+use futuresdr::blocks::FirBuilder;
+use futuresdr::blocks::Source;
+use futuresdr::runtime::Flowgraph;
+use futuresdr::runtime::Runtime;
+
+use futuredsp::firdes;
+use futuresdr::anyhow::Result;
+
+fn main() -> Result<()> {
+    let mut fg = Flowgraph::new();
+
+    const SAMPLING_FREQ: u32 = 44_100;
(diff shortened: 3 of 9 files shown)


bastibl:
Very nice, thank you! I'll add the `log` crate as a dependency in a follow-up commit and import `debug!`, `warn!`, etc macros. I guess this is fine for this library crate.

We could, at some point, experiment with adding more to the type system, for example (pseudo code):
```rust
/// A generic trait for filter window functions.
pub trait FilterWindow<TapsType, const N: usize> {
    /// Returns the window function at index `index`
    fn get(&self, index: usize) -> Option<TapsType>;

    /// Returns the number of taps.
    fn len(&self) -> usize {
        N
    }

    /// Returns the taps.
    fn taps(&self) -> Box<[TapsType; N]> {
        (0..N).map(|i| self.get(i).unwrap()).collect()
    }
}
```

I think this can make a performance difference, since it allows propagating the length of the taps array, which could lead to more efficient implementations. For example, if it knows that the number of taps are a multiple of 4 and SIMD always takes chunks of 4 samples, it doesn't have to worry about issues at end (i.e., an iteration with less than 4 samples).

Another thing that we could think of is implementing everything with a [generic float type](https://docs.rs/num/latest/num/trait.Float.html). But it might not be worth it. Or just create everything w/ `f64` and cast to `f32`? I'm not sure. Might be over-engineering.

Thanks again for the PR. Not having firdes functionality was a major pain point.

AndersKaloer:
Thanks, I agree that a generic float type might be the way to go, especially because the decision between `f32` and `f64` will be a recurring issue. However, I have not used `num::Float` before, so I do not know if it is the best solution to this.

Regarding the `FilterWindow` trait, I also agree that having a length will make the integration with SIMD, etc. easier. I did not think of that. Maybe having a `FilterWindow` trait is not needed at all, and we could simply use an array or a `Vec<T>`.

PR #48, opened 2022-03-24
Firdes refactoring

AndersKaloer:
This includes a few proposed changes to `firdes` and related modules:

- Window functions no longer use the `FilterWindow` trait, but directly returns the window as `Vec<f64>`. The use of `FilterWindow` trait seemed unnecessary, added a tight dependency between modules, and complicated the use of e.g., simd.

- Several of the window functions now includes a `periodic` argument that specifies whether the window should be constructed to be periodic. This is useful for spectral analysis (e.g., used in radar).

-  `firdes` now supports generic numeric types using the `num_traits::FromPrimitive` trait. This includes of course `f32` and `f64`, but can also be used for e.g., fixed-point numbers. Basically, everything is computed with `f64`, and then converted to the desired type before returning the filter taps.

- `firdes` now also includes a RRC filter.

In principle, the `num_traits::FromPrimitive` could also be used for the window functions, but I do not see the immediate benefit at the moment. It can be implemented later if needed.

### DIFF
--- a/futuredsp/src/firdes.rs
+++ b/futuredsp/src/firdes.rs
@@ -1,11 +1,10 @@
 //! Methods for designing FIR filters.
 
 extern crate alloc;
-use crate::windows::FilterWindow;
 use alloc::vec::Vec;
 
 /// Constructs a lowpass FIR filter with unit gain and cutoff frequency `cutoff` (in cycles/sample)
-/// using the specified window.
+/// using the specified window. The length of the filter equals the length of `window`.
 ///
 /// Example usage:
 /// ```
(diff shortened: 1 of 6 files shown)


futuredsp/src/firdes.rs
```diff
@@ -1,109 +1,106 @@
...
+    for (n, tap) in taps.iter_mut().enumerate() {
+        let x = n as f64 - alpha;
+        let filter_tap = match x == 0.0 {
             true => 1.0,
-            false => (omega_c * x).sin() / (core::f32::consts::PI * x),
+            false => (omega_c * x).sin() / (core::f64::consts::PI * x),
         };
-        taps.push(tap * window.get(n));
+        *tap *= filter_tap;
     }
```
lkolbly, line 38: If you want, I think this could also be something like:
```
window.iter().enumerate().map(|(n, tap)| {
        let x = n as f64 - alpha;
        let filter_tap = match x == 0.0 {
            true => 1.0,
            false => (omega_c * x).sin() / (core::f64::consts::PI * x),
        };
        tap * filter_tap;
}).map(|x| T::from_f64(*x).unwrap()).collect::<Vec<T>>()
```
to save a heap allocation.

lkolbly (approved):

futuredsp/src/firdes.rs
AndersKaloer: Thanks!

futuredsp/src/firdes.rs
```diff
@@ -1,109 +1,106 @@
...
+    let omega_c = 2.0 * core::f64::consts::PI * cutoff;
+    let alpha = (window.len() - 1) as f64 / 2.0;
+    window
+        .iter()
+        .enumerate()
+        .map(|(n, tap)| {
+            let x = n as f64 - alpha;
+            let filter_tap = match x == 0.0 {
+                true => 1.0,
+                false => (omega_c * x).sin() / (core::f64::consts::PI * x),
```
lkolbly, line 37: Okay, so this is unrelated to this PR, but since we're here...

I made a simple flowgraph like this:
```rust
fn main() -> Result<()> {

    let src = fg.add_block(futuresdr::blocks::Source::<Complex<f32>>::new(|| {
        let mut rng = rand::thread_rng();
        let x: f32 = rng.gen();
        let y: f32 = rng.gen();
        Complex {
            re: x * 2.0 - 1.0,
            im: y * 2.0 - 1.0,
        }
    }));

    let window = futuredsp::windows::BlackmanWindow::new(59);
    let taps = futuredsp::firdes::lowpass(59, 0.25, window);
    let filter = fg.add_block(FirBuilder::new::<Complex<f32>, f32, _>(taps));

    let sink = fg.add_block(FileSink::<Complex<f32>>::new("tmp.cf32"));

    fg.connect_stream(src, "out", filter, "in")?;
    fg.connect_stream(filter, "out", sink, "in")?;

    Runtime::new().run(fg)?;

    Ok(())
}
```
which, effectively, generates some complex white noise, runs it through a filter, and pipes that to a file which we can then analyze.

So, if I run this flowgraph as-is, I get a spectrum that looks something like:
image
with ~15dB of attenuation outside the passband.

If I change `omega_c` in `lowpass()` to `2.0 * cutoff` (removing the π factor) and change the "false" tap calculation to `(omega_c * x).sin() / (omega_c * x)` (from `(omega_c * x).sin() / (core::f32::consts::PI * x)`), so that the overall calculation looks like:
```rust
    let omega_c = 2.0 * cutoff;
    let alpha = (num_taps - 1) as f32 / 2.0;
    for n in 0..num_taps {
        let x = n as f32 - alpha;
        let tap = match x == 0.0 {
            true => 1.0,
            false => (omega_c * x).sin() / (omega_c * x),
        };
        taps.push(tap * window.get(n));
    }
```
I get:
image
Obviously, the passband is a different width, but more notably the attenuation is higher outside the passband.

I know essentially nothing about signal processing (I stole my modifications from [here](https://tomroelandts.com/articles/how-to-create-a-simple-low-pass-filter)). I also haven't used the `firdes::kaiser` methods yet. But, is this known/expected behaviour?

AndersKaloer: You are right that this is not how it is supposed to look.
I had a closer look, at it seems that the error is that the "true" tap should not be `1.0` but `omega_c / core::f32::consts::PI`. Changing that seems to solve the problem for me. I will provide a fix.

In fact, the implementation that we use is the same as the sinc filter in the link you provided:
```
h[n] = 2*fc*sinc(2*fc*n)
     = 2fc*sin(pi*2*fc*n) / (pi*2*fc*n)
     = sin(pi*2*fc*n) / (pi*n)
     = sin(omega_c*n) / (pi*n)
```
where `omega_c = pi*2*fc`. Since `sinc(0) = 1` we have to use `h[0] = 2*fc = omega_c / pi` (and obviously not `h[0] = 1.0`).

lkolbly: Ah okay, that derivation makes sense to me. But yeah replacing the "true" tap worked on my end too, thanks!

AndersKaloer:
It might be possible, but I guess it requires the number of taps be known at compile time. If for instance the kaiser method is used to design the filter, then the number of taps depends on the transition width and ripple. In practice, I guess these parameters will often be static, and so the `design_kaiser_window()` function can be made `const`. The question is whether the benefits of using a static filter length outweigh these constraints. There might be a way to achieve both flexibility, but I am not sufficiently proficient in Rust to see that option :)

Issue #50, opened 2022-04-16
Option to change soapy device properties at runtime


JuliDi:
Hey,

Is it possible to change soapy device properties like the center frequency etc. at runtime?  Or more general: Can the flowgraph be changed at runtime?

I would like to avoid stopping the flowgraph to create a new `SoapySource` object, which will cause a new device initialisation that takes a few seconds.
And if so, how could that be implemented in FutureSDR? I did not find a way to access the device after it is converted to a block.

From what I see in the soapy crate doc (https://kevinmehall.net/rustdoc/soapysdr/soapysdr/struct.Device.html#method.frequency) this _might_ be possible, but I am anything but an expert on this.

What I would like to do at the end is an extension of the `fm-receiver` example:

```rust
let src = SoapySourceBuilder::new()
        .freq(93.7e6)
        .sample_rate(3.2e6)
        .gain(40.0)
        .build();

// Connect everything ...

// Execute 
Runtime::new().run(fg)?;
// as an asynchronous task/thread/whatever

// Make changes to the frequency to tune e.g. to a specific radio station on the go
```

bastibl:
The Soapy Source already has a basic handler to change the frequency: https://github.com/FutureSDR/FutureSDR/blob/master/src/blocks/soapy_src.rs#L36-L55 This could be extended to accept other types of PMTs. And one could add more handlers in a similar fashion.

Instead of `run`ning the flowgraph, which awaits the task until the flowgraph is finished. You can [`start`](https://docs.rs/futuresdr/latest/futuresdr/runtime/struct.Runtime.html#method.start) it, which gives you a [FlowgraphHandle](https://docs.rs/futuresdr/latest/futuresdr/runtime/struct.FlowgraphHandle.html). With the handle you can call message handlers of blocks. Alternatively, you can use the REST API to issue callbacks.

JuliDi:
Thanks for the input, I will give that a try!

PR #51, opened 2022-05-07
Expand fm-receiver example with SoapySDR freq change at runtime

JuliDi:
Following up on #50, this PR expands the fm-receiver example and adds some documentation. Now the user can change the frequency at runtime from the command line.
I found it a good example to get started with FutureSDR since it is both simple and makes use of many interesting features of the crate.

Moreover, one can now the sample rate for a `SoapySource` block at runtime, as well (not used in the fm-receiver example), and there is some minimal documentation for the block.

Regarding tests:
The code is tested and confirmed working with a LimeSDR Mini. It seems difficult to add automated tests for the SoapyBlock – presumably a device would have to be emulated for this. If you have a solution for this, please let me know. I'd be happy to add a few test cases.


bastibl:
Cool, thanks for the extension!

I agree, the FM receiver is the *hello world* of SDR, so the example should be nice, well documented, and work well on all platforms. I think, at the moment, the last remaining issue is the lack of a resampler, since many audio stacks are not so flexible with regards to supported input rates. I hope, I have some time to look into this soon.

Another thing is the need for the sleep. That's not nice :-/ Ideally, the `fg.start` would return only after all blocks are initialized (i.e., all of them are done with their `init` function). Another thing, I need to look into. So some homework for me :-)

Regarding the changes to `.gitignore`. I'm not a big fan of adding this to projects. Ideally, this should go into the users global gitignore file. Because, if tomorrow someone releases a new editor, it shouldn't be required to update all `.gitignore` files of all projects. Instead, the users of this editor should update their global (user-wide) ignore file.

JuliDi:
Regarding the resampling: I thought about adding the possibility to change the sample rate, as well, but that would also require the audio sink to change the sample rate at runtime (which is not yet possible afaik).
Maybe the DSP part could offer a decimation filter? There is already an FIR filter block, so adding some low-pass filtering and a downsampler seems rather straightforward. One could even make it just two blocks, actually.
I will look into this when I have the time.

> Ideally, the fg.start would return only after all blocks are initialized

That would be great 

>Ideally, this should go into the users global gitignore file

Good idea, thanks for the input! I was not aware that this would work, the git global settings are something I am struggling with every time I change something.

bastibl:
I think the `AudioSink` should provide a `AudioSink::supported_rates() -> Vec<u64>` method. Then one can select one of those when constructing the block and put the corresponding resampler in-between SDR and audio sink. A simple resampler is probably not that hard, but I didn't have a look yet.

Issue #49, opened 2022-04-06
AudioSink add complex input or multiple taps


ratzrattillo:
When using the AudioSink block with two channels, i would expect the block to have two input taps (e.g. one for the left and one for the right channel).
@loic-fejoz suggested two have either one tap per channel, or one tap accepting a Complex number as input.

Currently the AudioSink in blocks/audio/audio_sink.rs:40 creates the stream with a hardcoded size of 4 Bytes per sample:

`StreamIoBuilder::new().add_input("in", 4).build(),`

To play an audio file with only one channel (like the rick.mp3 example) on a windows PC, which in many cases only show an audio output device with two channels, hacks like this have to be done:

`
use cpal::traits::{DeviceTrait, HostTrait};
use futuresdr::anyhow::Result;
use futuresdr::blocks::ApplyIntoIter;
use futuresdr::blocks::audio::AudioSink;
use futuresdr::blocks::audio::FileSource;
use futuresdr::runtime::{Flowgraph, Runtime};

fn main() -> Result<()> {

    let host = cpal::default_host();
    let audio_device = host.default_output_device().expect("no output device available");
    println!("Device name: {}", audio_device.name().unwrap());
    let supported_configs_range = audio_device.supported_output_configs().expect("error while querying configs");
    for format in supported_configs_range {
        println!("Sample Rate: Min: {:?}, Max: {:?}", format.min_sample_rate(), format.max_sample_rate());
        println!("Channels: {}", format.channels());
        println!("SampleFormat: {:?}", format.sample_format());
    }

    let filename = "rick.mp3";
    let mut fg = Flowgraph::new();

    let src = FileSource::new(filename);
    let inner = src.as_async::<FileSource>().unwrap();

    fn tuple(s: &f32) -> Vec<f32> {
        let mut vec = Vec::with_capacity(2);
        vec.push(*s); // L
        vec.push(*s); // R
        return vec;
    }
    let custom = ApplyIntoIter::<f32, Vec<f32>>::new(tuple);
    let snk = AudioSink::new(inner.sample_rate(), 2); // inner.channels()

    println!("File: {}, Sample Rate: {}, Channels: {}", filename, inner.sample_rate(), inner.channels());

    let src = fg.add_block(src);
    let custom = fg.add_block(custom);
    let snk = fg.add_block(snk);

    fg.connect_stream(src, "out", custom, "in")?;
    fg.connect_stream(custom, "out", snk, "in")?;

    let rt = Runtime::new();
[...]

PR #52, opened 2022-05-14
feat(blocks): Introduce ApplyNM block

loic-fejoz:
The ApplyNM block applies a function on N input samples  and ouput M samples. It enables autovectorization.
It is very handy for interleaved samples for instance.
Somehow related to discussion in #49


bastibl:
I cherry-picked the commits. GitHub doesn't get it, but it's in :-)

PR #53, opened 2022-05-30
Fix fir filter convolution

AndersKaloer:
The current FIR filter implementation does not reverse the filter taps before performing the convolution, leading to a correlation operation rather than convolution. This pull request fixes the problem.

For instance, passing a signal `[1.0, 2.0, 3.0, 4.0, 5.0, 6.0]` through a filter with impulse response `[1.0, 2.0]` should produce `[4.0, 7.0, 10.0, 13.0, 16.0]`, but the current implementation produces `[5.0, 8.0, 11.0, 14.0, 17.0]`.

It can be verified with the following example:
```rust
use futuresdr::anyhow::{Context, Result};
use futuresdr::blocks::FirBuilder;
use futuresdr::blocks::VectorSink;
use futuresdr::blocks::VectorSinkBuilder;
use futuresdr::blocks::VectorSourceBuilder;
use futuresdr::runtime::Flowgraph;
use futuresdr::runtime::Runtime;

fn main() -> Result<()> {

    let input = vec![1.0_f32, 2.0, 3.0, 4.0, 5.0, 6.0];
    let src = fg.add_block(VectorSourceBuilder::new(input.clone()).build());
    let filter = fg.add_block(FirBuilder::new::<f32, f32, _>([1.0, 2.0]));
    let snk = fg.add_block(
        VectorSinkBuilder::<f32>::new()
            .init_capacity(input.len())
            .build(),
    );

    fg.connect_stream(src, "out", filter, "in")?;
    fg.connect_stream(filter, "out", snk, "in")?;

    fg = Runtime::new().run(fg)?;

    let snk = fg
        .kernel::<VectorSink<f32>>(snk)
        .context("block not found")?;
    let v = snk.items();
    println!("{:?}", v);

    Ok(())
}
```

### DIFF
--- a/futuredsp/src/fir.rs
+++ b/futuredsp/src/fir.rs
@@ -68,7 +68,11 @@ where
         for k in 0..n {
             let mut sum = init();
             for t in 0..taps.num_taps() {
-                sum = mac(sum, *i.get_unchecked(k + t), taps.get(t));
+                sum = mac(
+                    sum,
+                    *i.get_unchecked(k + t),
+                    taps.get(taps.num_taps() - 1 - t),
+                );
             }
             *o.get_unchecked_mut(k) = sum;
         }
@@ -165,7 +169,7 @@ mod tests {
             kernel.work(&input, &mut output),
             (1, 1, ComputationStatus::InsufficientInput)
         );
-        assert_eq!(output[0], 14.0);
+        assert_eq!(output[0], 10.0);
 
         let mut output = [];
         assert_eq!(
@@ -178,7 +182,7 @@ mod tests {
             kernel.work(&input, &mut output),
             (1, 1, ComputationStatus::InsufficientInput)
         );
-        assert_eq!(output[0], 14.0);
+        assert_eq!(output[0], 10.0);
 
         let input = [1.0, 2.0, 3.0, 4.0, 5.0];
         let mut output = [0.0; 2];
@@ -186,8 +190,8 @@ mod tests {
             kernel.work(&input, &mut output),
             (2, 2, ComputationStatus::InsufficientOutput)
         );
-        assert_eq!(output[0], 14.0);
-        assert_eq!(output[1], 20.0);
+        assert_eq!(output[0], 10.0);
+        assert_eq!(output[1], 16.0);
     }
 
     /// Tests the "terminating condition" where the input is finished and the


bastibl:
Thank you. My understanding is that this doesn't matter for `firdes`, since they are symmetric anyway, no?

Also, is this the way you would, intuitively, expect the API to work or should we maybe reverse the taps when creating the FIR block and pass the reversed taps to the FIR? Would probably also be faster (I assume). Or do you think this would be a strange API for the FIR function in `futuresdr`?

AndersKaloer:
The problem arises with the polyphase decompositions, which are generally not symmetric.

The PR is indeed a bit slower than the current implementation. As you say, we could specify reversed taps (and I could reverse when I decompose the filter into polyphase components), but I find it a bit odd - at least it should be documented :)

I think the most intuitive way would be to reverse the taps in e.g., `PolyphaseResamplingFirKernel::new()`, but it is a bit tricky with the `TapsAccessor` trait. We could add a `TapsAccessor::reverse()` method or similar, but it would require the types to either be mutable or `TapsAccessor::reverse()` to return a new instance.

Adding a `reverse()` method would be relatively simple, but the question is how many methods we want to add to `TapsAccessor`. For instance, I would also prefer to split the taps in the polyphase filter (#54) into multiple separate `TapAccessor`s, but this would require yet another method to the `TapAccessor` trait (or we should do the actual polyphase decomposition in `firdes` instead of `PolyphaseResamplingFirKernel`).

In the end, I guess it is a matter of preference and where to put the complexity. Performance wise, there is no doubt that we should aim for aligned arrays.

bastibl:
Thanks for your reply. I didn't consider the interaction with the `TapAccessor`. So let's go with this approach.

AndersKaloer:
JuliDi, I agree that it would probably be more efficient.

I guess `TapsAccessor` needs to be modified to require also the `IntoIterator`. That could probably be quite useful in many situations. 
To support `.rev()` as in your example, I believe we would also have to ensure that `IntoIterator::IntoIter` is a `DoubleEndedIterator`.

Do you think this could be done in a clean way without requiring `where` clauses everywhere?

JuliDi:
Thanks for your Input, @AndersKaloer!
Presumably, without some tricks this could get very nasty indeed. There seem to be ways to implement traits for multiple types at once using macros (see e.g. https://stackoverflow.com/questions/39150216/implementing-a-trait-for-multiple-types-at-once).
For example, the implementation of the `get()` function for the `TapsAccessor` is the same for all types.https://github.com/FutureSDR/FutureSDR/blob/a92aed12a74b7978bb45bb5be43f9ebe0a1930b7/futuredsp/src/tapsaccessor.rs#L13
For the `num_taps()` on the arrays, I don't see the exact reason for not using `self.len()`. Does that not work here?

***

I've just come across this page https://doc.rust-lang.org/1.30.0/book/second-edition/ch19-03-advanced-traits.html#using-supertraits-to-require-one-traits-functionality-within-another-trait and started wondering whether we could just have the `TapsAccessor` trait require the `DoubleEndedIterator`. Then we automatically have that functionality – and for types that don't provide it by default, we can still implement it manually.

```Rust
trait TapsAccessor: Send, TapsAccessor { ... }
```

Am I overlooking something or might this actually work?

PR #54, opened 2022-05-30
Add polyphase rational resampler

AndersKaloer:
This pull request adds a polyphase implementation of a FIR rational resampler.

Specifically, it adds a `PolyphaseResamplingFirKernel` to `futuredsp::fir`, which implements a simple, stateless polyphase interpolator/decimator.

The following function in `futuredsp::firdes::kaiser` is provided to design Nyquist interpolation/decimation filters (similar to `designMultirateFIR` in Matlab):
```rust
pub fn multirate<T: FromPrimitive>(
        interp: usize,
        decim: usize,
        half_polyphase_len: usize,
        max_ripple: f64,
    ) -> Vec<T>
```

In addition, two functions are provided in `futuresdr::blocks::FirBuilder` to construct resampling blocks:
```rust
pub fn new_resampling<SampleType>(interp: usize, decim: usize) -> Block
```
which creates a interpolation/decimation filter using some standard parameters, and
```rust
    pub fn new_resampling_with_taps<SampleType, TapType, Taps>(
        interp: usize,
        decim: usize,
        taps: Taps,
    ) -> Block
```

Example usage can be found in `examples/firdes`.
I also have a working modification of `examples/fm-receiver` that uses the resampler decimate the high sampling rate before demodulation. Should we include that as well?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -102,6 +102,8 @@ futuresdr-pmt = { path = "pmt", version = "0.0.3" }
 log = { version = "0.4", features = ["std", "max_level_debug", "release_max_level_info"] }
 lttng-ust = { version = "0.1.0", optional = true}
 num-complex = "0.4.0"
+num-traits = "0.2"
+num-integer = "0.1"
 num_cpus = "1.13.0"
 once_cell = "1.5.2"
 rand = "0.8.0"
--- a/examples/firdes/src/main.rs
+++ b/examples/firdes/src/main.rs
@@ -10,7 +10,7 @@ use futuresdr::anyhow::Result;
 fn main() -> Result<()> {
     let mut fg = Flowgraph::new();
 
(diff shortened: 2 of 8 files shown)


bastibl:
Awesome! Seeing the notification about this PR in my inbox in the morning really made my day :-) I'll probably have some time to look at it on Friday. Your modification of the fm-receiver would be very much appreciated. Not having a resample was a recurring issue for this example.

AndersKaloer:
Excellent, I added the FM receiver example.
Also, let me know if you think the API could be designed in a better way.

bastibl:
Tbh, I don't know much about the DSP bits, but it looks very good to me. Also thanks for the documentation and tests -- very much appreciated. I'll merge it and will do some testing on top. Now we should be able to extend the fm-receiver and similar apps to query for the default sample rate of the audio device and resample accordingly.

PR #63, opened 2022-09-01
Flowgraph as mermaid diagram

loic-fejoz:
Add capability to display flow-graph as a [mermaid diagram](https://mermaid-js.github.io/mermaid/#/) available at http://localhost:26125/main/graph/

It is rendered as such:

This pull-request also add an example to decode SSB (actually upper-side band). It illustrates how to set name of block in your flow-graph.

NB: the route is under `/main` because eventually I would like to add other http route handlers, either for http audio stream or more applicative one.

The conversion could also be called directly on a flowgraph as in the CW example:
```rust
fg.to_mermaid(std::io::stdout());
```

This PR also bring file source repetition as used in the SSB example.


bastibl:
Thanks for the PR! As discussed, we close this in favor of the cherry-picked, upstreamed changes and this PR https://github.com/bastibl/FutureSDR/pull/1.

PR #64, opened 2022-09-17
Channel source

ratzrattillo:


bastibl:
This is a very cool idea! One thing, I wondered is whether this should be a channel of something like `Box<[T]>`, i.e., where one can push chunks of samples. I don't know if you had a particular use-case in mind, but I guess most of the time, the code at the sending side of the channel has samples in (large) chunks. The current version would require to push each sample individually, which might be inefficient.

ratzrattillo:
Fixed it!

bastibl:
Great, thank you! I'll rebase and do some follow-up commit to add a test and fix a minor issue with small vectors. Hope that's fine for you. I don't want to annoy you with a lot of back-and-forth. Let me know if you prefer otherwise.

PR #65, opened 2022-10-01
feat(blocks): Selector

loic-fejoz:
Add `Selector<_,N,M>` block that forward the input at given index to the output at given index.
Think of it as a selectable switch.

Add an audio example `play-selectable-tone` for illustration purpose of Selector.
The drop policy is configurable for new comer to test and understand the impact of it.


bastibl:
I cherry-picked the commits and did a follow-up to adapt for recent changes. Looks like GitHub doesn't recognize it, but they are in. Thank you!

Issue #62, opened 2022-08-20
Only first TX/RX port can be used on soapy+usrp b210?


astrojhgu:
https://github.com/FutureSDR/FutureSDR/blob/77f1a44f59847be85593eed3900b0294e44ea3eb/src/blocks/soapy_src.rs#L146

When getting stream from `soapydev`, the channel is hardcoded to 0.

bastibl:
At this stage, the project is unfortunately full of those limitations. I didn't have the time to create a full-fledged block library but usually extend blocks as needed. Any help is greatly appreciated. Feel free to ask questions here or on Discord, if you want to give it a shot.

etschneider:
I'm working on multi-channel support. It's currently working, but needs a little more polish. If you'd like to try it out, see:

https://github.com/etschneider/FutureSDR/tree/ets_soapy_multichan

Issue #70, opened 2022-10-05
Make SoapySource/Sink new() parameters optional


etschneider:
Using the `dev` parameter allows for flexibility in preconfiguring a Soapy device, but the `freq`,`sample_rate`, and `gain` parameters are always applied, which may override the desired device settings.

I propose making those inputs `Option<>`.

A PR will follow shortly.

There are some other architecture questions I have about the Soapy implementation (and block/arch in general), perhaps the discussion area (Q&A?) might be the right place for that? 

For example, is `Kernel::init()` the mandated place for all device init, as opposed to doing it during creation of the block. It seems cleaner to me to init the device (and stream) in `new()` and just do the `stream.activate()` in `Kernel.init()`. That would alleviate the need to carry the initialization params in the sink/source struct, only to be used during flowgraph initialization.

bastibl:
Thanks for the PR!

Regarding `init`: I agree that the code would be nicer, if this was done during creation of the block. However, in the future, it should be possible to stop, reconfigure, and restart a flowgraph. In that case, `init` would be run every time the flowgraph is (re)started. At least that was the idea when adding it.

Also Soapy and the underlying drivers might have uncommon constraints and `init` is run in the context of the thread that's running `work`. In case, a driver uses thread local storage, it would break things, if we opened in the constructor. (Thinking about this, that's yet another thing that the `dev` parameter breaks :-)).

Maybe, we should double-down on the builder pattern? Then we could keep the builders around and just create new blocks when we restart flowgraph?

As mentioned on Discord, my plan was anyhow to have serializable builders to allow straightforward flowgraph instantiation on remote hosts.

PR #72, opened 2022-10-06
Soapy multichan

etschneider:
The implementation should be backwards compatible via the builder interface. The new() signature changes the chan param to a Vec.
    
This builds upon the optional parameters PR #71.
   
Fixes: #62
Refs: #70, #71


bastibl:
I cherry-picked the two commits that are on top of #71 to avoid the merge commit. Thank you!

PR #73, opened 2022-10-07
zigbee improvements

meinscheins:
added parameters for the zigbee rx and tx examples and added a udp sink for the rx example similar to the rx wlan example


bastibl:
thanks! :-)

PR #74, opened 2022-10-08
pmt: add VecPmt and MapStrPmt

etschneider:
Limited HashMap to String keys since the Hash trait on all Pmt types would be a burden (esp PmtAny).


bastibl:
Thanks!

PR #75, opened 2022-10-08
Ets pmt serde

etschneider:
Putting it here for review, I expect there will be some questions/comments/discussion.

### DIFF
--- a/pmt/Cargo.toml
+++ b/pmt/Cargo.toml
@@ -13,6 +13,8 @@ categories = ["asynchronous", "concurrency", "hardware-support", "science", "was
 [dependencies]
 dyn-clone = "1.0.9"
 serde = { version = "1.0", features = ["derive"] }
+typetag = "0.2"
 
 [dev-dependencies]
 flexbuffers = "2.0.0"
+serde_json = "1.0"
--- a/pmt/src/lib.rs
+++ b/pmt/src/lib.rs
@@ -107,6 +107,9 @@ impl PartialEq for Pmt {
             (Pmt::VecF32(x), Pmt::VecF32(y)) => x == y,
             (Pmt::VecU64(x), Pmt::VecU64(y)) => x == y,
             (Pmt::Blob(x), Pmt::Blob(y)) => x == y,
+            (Pmt::VecPmt(x), Pmt::VecPmt(y)) => x == y,
+            (Pmt::MapStrPmt(x), Pmt::MapStrPmt(y)) => x == y,
+            //How to handle Any?
             _ => false,
         }
     }
@@ -266,6 +269,10 @@ mod test {
         let f3 = Pmt::F32(0.2);
         assert_eq!(f1, f2);
         assert_ne!(f1, f3);
+
+        // How to handle this?
+        // let pa = Pmt::Any(Box::new((1, 2, 3)));
+        // assert_eq!(pa, pa);
     }
 
     #[test]
--- a/pmt/src/lib.rs
+++ b/pmt/src/lib.rs
@@ -2,37 +2,76 @@ use dyn_clone::DynClone;
 use serde::{Deserialize, Serialize};
 use std::any::Any;
 use std::collections::HashMap;
+use std::fmt;
 
 mod description;
 pub use description::BlockDescription;
 pub use description::FlowgraphDescription;
 
-#[typetag::serde(tag = "type")]
-pub trait PmtAny: Any + DynClone + Send + Sync + std::fmt::Debug + 'static {
+pub trait PmtAny: Any + DynClone + Send + Sync + 'static {
     fn as_any(&self) -> &dyn Any;
     fn as_any_mut(&mut self) -> &mut dyn Any;
 }
 dyn_clone::clone_trait_object!(PmtAny);
 
-// Unfortunately, typetag (v0.2) does not support automatic deserialization via
-// #[typetag::serde] on generic impls, so we need to impl the any functions
-// for every instance.
-//
-// #[typetag::serde]
-// impl<T: Any + DynClone + Send + Sync + 'static> PmtAny for T {
-//     fn as_any(&self) -> &dyn Any {
-//         self
-//     }
-//     fn as_any_mut(&mut self) -> &mut dyn Any {
-//         self
-//     }
-// }
+impl<T: Any + DynClone + Send + Sync + 'static> PmtAny for T {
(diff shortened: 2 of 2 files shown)


bastibl:
Thanks for the PR! I think, it would be nice to allow serialization of `dyn Any` types. However, as we already discussed, I think we probably shouldn't force it on all types. Instead, we could have this as an additional `Pmt` type.

This commit, for example, breaks the WLAN receiver and there are no straightforward ways to fix, since it uses base types, like `(String, u32)` and these couldn't be extended to implement `PmtAny`. The WLAN transceiver (or FutureSDR) cannot implement `PmtAny` for `(String, u32)` since neither the trait nor the data structure are from that crate.

```
error[E0117]: only traits defined in the current crate can be implemented for arbitrary types
 --> src/main.rs:5:1
  |
5 | impl PmtAny for (String, u32) {
  | ^^^^^^^^^^^^^^^^-------------
  | |               |
  | |               this is not defined in the current crate because tuples are always foreign
  | impl doesn't use only types from inside the current crate
```
I think we should try to get a better understanding about the issue with generic implementations. Is this a problem of the `typetag` macro/crate or is this a technical issue of Rust? Maybe there is a way to manually implement `PmtAny` for all types.

bastibl:
(Btw, you don't have to merge back `main`. If you want to keep the PR up-to-date, a rebase would be preferred, since I think it's nicer to not have merge commits in the `main` branch.)

etschneider:
Yes, the single `Any` was an intentional discussion provoker. :-)

The foreign types case is, AFAIK, a pretty common Rust issue that is normally addressed via a wrapping struct, e.g.:
```
// Not too onerous...
#[derive(Clone, Debug, Serialize, Deserialize)]
struct MyStruct( (String, u32) );
        
// This whole bit could be a one line macro invocation.
#[typetag::serde]
impl PmtAny for MyStruct {
    fn as_any(&self) -> &dyn Any { self }
    fn as_any_mut(&mut self) -> &mut dyn Any { self }
}

// Ok...
let p = Pmt::Any(Box::new(MyStruct( ("hi".to_owned(), 123) )));
```
I'll need to look at WLAN for a concrete example though.

I am probably missing the point, but to use Pmt we have to put our something in a Pmt anyway, so some wrapping is already happening, so a layer around whatever foreign types you need in a PmtAny doesn't seem that out of place.

My main argument for the single PmtAny type is that the vast majority of cases would want serialization, the implementation is trivial for most cases with a derive, and the few cases that are problems for serde can be handled as needed via serde ignores or dummy implementations.

The alternative is to have two PmtAny types to go along with the enum options. I'm not yet convinced that there is a case that could not be made to fit into the single serializable Any. (But I will go review WLAN to see what I'm missing).

I figured you might have an issue with the Debug trait too. But having debugs available is pretty nice IMO. :-)

Re: typetag generics, I don't know, but I'm guessing it is a technical issue, since the crate is from the Serde author. But really that only means that PmtAny users have to add their own PmtAny impl (which can be done via macro). It is a shortcoming, but a pretty minor one IMO.

etschneider:
Ok, `Pmt::Any` and `Pmt::AnySerde` now.

bastibl:
Thanks for the change. With this, it could be merged without forcing it on everybody.

The wrapper struct, for example, is easy, but it's not obvious for the user, why it should be required. Also the `Any` implementation is easy, but it's not clear to the user that this is required, etc. What I mean is that, if we would add more and more of such things in the code base, we would just drive everybody crazy.

I think, we should add a derive proc macro for it. E.g.:

```rust
#[derive(Clone, Debug, Pmt)]
struct MyStruct( (String, u32) );
```
or `PmtSerde` or something.

etschneider:
Yeah, it is probably the right call at this time. 

My chief concern is managing combinatorial explosion in the future. What happens when there are a few more traits that would be really great to have? Do we implement a separate type for every possible combination of traits? 

Debug as kind of silly example. Should there be a `PmtAny`, `PmtSerde`, `PmtAnyDebug`, `PmtAnySerdeDebug`? Or do we just make it an expected trait? Adding a debug trait is trivial in most cases. Just trying to think ahead, choices early tend to be hard to change later.

On that note, do you think I should remove the debug trait from `PmtAnySerde`? :-) Your call.

What about `PartialEq`? Currently PmtAny is not going to play nice.

bastibl:
Tbh, I don't understand the issue with `Debug`, since the initial information on the PR was quite minimal. I also don't know what you are referring to with the issue about `ParitalEq`. Requiring many traits is OK, but in my opinion, they shouldn't be a big burden, i.e., they should be easy to derive with a proc macro.

Since this doesn't exist, I'm hesitant to force it on all types. With a separate option, it shouldn't hurt anyone. It doesn't mean that his has to stay that way forever (and that we will experience this bloat). If this gets more stream-lined in the future. And if we figure that a proc macro works for all (or close all) cases. Then we could drop the non-serde case in the future, since then migration is trivial. But you didn't do that yet.

As much as I wanted, I don't have the time at the moment to experiment with the API and explore all implications myself. And with the replies here, I'm only more puzzled -.- If this is helpful for your project, we can merge it.

etschneider:
I agree entirely with your thoughts. I suggest merging with the two options, but I'm in no hurry if you prefer to wait.

Re `PartialEq`:
```
let a = Pmt::Any(Box::new((1, 2, 3)));
assert_eq!(a, a);
```
fails due to to the `_ => false` in `impl PartialEq for Pmt` .

I was mistakenly thinking we could add `PartialEq` to the `PmtAny` trait bounds, but that isn't true, so I don't know how to overcome that failure case.

bastibl:
Just wanted to merge this, but it looks like `typetag` cannot be compiled to wasm :-(
```
error: #[ctor] is not supported on the current target
   --> pmt/src/lib.rs:220:9
    |
220 |         #[typetag::serde]
    |         ^^^^^^^^^^^^^^^^^
    |
    = note: this error originates in the attribute macro `$crate::ctor` which comes from the expansion of the attribute macro `typetag::serde` (in Nightly builds, run with -Z macro-backtrace for more info)
```

See https://github.com/dtolnay/typetag/pull/36

etschneider:
Welp, that is a bummer! I guess it is back to the drawing board for solutions...

I really like the Pmt::Any, but serialization would really make it much more useful. I'll give this some thought to see if I come up with any ideas. Per the typetag reference you provided, it does seem like we will need some sort of runtime type registration which would make the usage a bit less ergonomic.

Issue #80, opened 2022-10-10
Do not return an error when flowgraph was terminated exteranlly.


bastibl:
There is currently no differentiation between termination because of a block error and external termination of the flowgraph through the flowgraph handle. In both cases the flowgraph task returns an `Err("Flowgraph was terminated")`. While this is fine when a block failed, we should probably try to shutdown gracefully in case of external termination. Maybe we should introduce `Kill` (in case of block error) and `Shutdown` (from external). In the latter case, the flowgraph task should succeed. See this example: https://github.com/samcarey/drop_block_test/blob/main/src/main.rs#L20

etschneider:
I've been looking into this. The direct fix is easy, just remove the `bail!` in `run_flowgraph`/`FlowgraphMessage::Terminate`, all the blocks get a chance to terminate and the loop exits on the `active_blocks == 0` condition. That seems like the right behavior.

Perhaps there should be a a time limit to wait on the blocks to terminate (in case one fails to terminate correctly).

However, the provided example has different problem. The provided code deadlocks, but not because of the runtime result. 

Interestingly, the `tests/runtime::fg_terminate` does not deadlock, even when checking the `task.await` result. The workaround is simple, but not intuitive (at least to me).

This deadlocks:
```rust
let (task, mut fg_handle) = block_on(Runtime::new().start(fg));
```

This works:
```rust
let rt = Runtime::new();
let (task, mut fg_handle) = block_on(rt.start(fg));
```

samcarey:
@etschneider's workaround seems to fix the deadlock for me, and the `deinit` and `drop` functions for the block get called too. Thanks! 

I decided to go ahead and wrap the task and handle in a struct that terminates from it's `drop` method so that cleanup will happen automatically, even if there's a panic. So I think my use case is addressed now!

```rust
struct RunningFlowgraph {
    task: Option<Task<Result<Flowgraph>>>,
    handle: FlowgraphHandle,
}

impl From<(Task<Result<Flowgraph>>, FlowgraphHandle)> for RunningFlowgraph {
    fn from((task, handle): (Task<Result<Flowgraph>>, FlowgraphHandle)) -> Self {
        Self {
            task: Some(task),
            handle,
        }
    }
}

impl Drop for RunningFlowgraph {
    fn drop(&mut self) {
        block_on(self.handle.terminate()).unwrap();
        block_on(self.task.take().unwrap()).err();
    }
}

fn main() {
[...]

PR #81, opened 2022-10-11
runtime:   do not return error on terminate msg

etschneider:
Also added an Err check in the executor loop of SmolScheduler::new() when testing. Probably should remain in some form.

Left commented-out deadlock code in:
tests/runtime::fg_terminate()

refs: #80

### DIFF
--- a/src/blocks/bad_block.rs
+++ b/src/blocks/bad_block.rs
@@ -1,0 +1,91 @@
+use std::{cmp, marker::PhantomData, ptr};
+
+use crate::{
+    anyhow::{bail, Result},
+    runtime::{
+        Block, BlockMeta, BlockMetaBuilder, Kernel, MessageIo, MessageIoBuilder, StreamIo,
+        StreamIoBuilder, WorkIo,
+    },
+};
+
+pub enum FailType {
+    Panic,
+    Error,
+}
+
+/// Intentionally generate errors to test the runtime.
+#[derive(Default)]
+pub struct BadBlock<T> {
+    pub work_fail: Option<FailType>,
+    pub drop_fail: Option<FailType>,
+    _phantom: PhantomData<T>,
+}
+
+impl<T: Clone + std::fmt::Debug + Send + Sync + 'static> BadBlock<T> {
+    pub fn to_block(self) -> Block {
+        Block::new(
+            BlockMetaBuilder::new("BadBlock").build(),
+            StreamIoBuilder::new()
+                .add_input::<T>("in")
+                .add_output::<T>("out")
+                .build(),
+            MessageIoBuilder::<Self>::new().build(),
+            self,
+        )
+    }
+}
+
+#[doc(hidden)]
+#[async_trait]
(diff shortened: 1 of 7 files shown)


etschneider:
As an extra note, a test log:

```
FutureSDR: DEBUG - in run_flowgraph
FutureSDR: DEBUG - connect stream io
FutureSDR: DEBUG - connect message io
FutureSDR: DEBUG - init blocks
FutureSDR: DEBUG - wait for blocks init
FutureSDR: DEBUG - running blocks
FutureSDR: DEBUG - Listening on 127.0.0.1:1337
FutureSDR: DEBUG - FlowgraphMessage::Terminate
FutureSDR: DEBUG - NullSink_0 terminating 
FutureSDR: DEBUG - NullSource_0 terminating 
FutureSDR: DEBUG - Throttle_0 terminating 
FutureSDR: DEBUG - No active blocks, exiting.
test fg_terminate ... ok
```

bastibl:
I think that, at the moment, we could get the `Terminate` message either from an external source through the flowgraph handle or when a block encounters an error, e.g., during work: https://github.com/FutureSDR/FutureSDR/blob/main/src/runtime/runtime.rs#L561-L567. My understanding is that this solution would only work in the former case, which is why I was suggesting the differentiation between the cases in #80. I thought that a block that fails wouldn't send the `Done` message, required to count down the active blocks. Or maybe I'm missing something.

etschneider:
This sounds like it should be part of the more general discussion of flowgraph lifecycles. I'll add comments on #80 for now.

etschneider:
Ok... I think I'm finally getting your point. So in `runtime::run_block()` change error conditions from:

```rust
main_inbox.send(FlowgraphMessage::Terminate).await?;
```
to:
```rust
main_inbox.send(FlowgraphMessage::Kill).await?;
```

Then in `run_flowgraph()` handle the `Kill` msg differently than `Terminate`?

If I'm on the right track here I can certainly make those changes.

Perhaps instead of kill, something like `FlowgraphMessage::BlockError` might be more descriptive?

But what is the different behavior aside from the return value? 

I think all blocks should still  be given the opportunity to exit gracefully if they can. Even the block that sent the `Kill`. 

I do think there should be a timer set to force exit if some blocks do not reply with `BlockDone`. Perhaps a general periodic `FlowgraphMessage::Tick` would be useful for that (and more). That would ensure that the msg handler gets activated periodically.

Might be useful as a `BlockMessage` too. `Kernel::start()/stop()/tick()` (#86, /me ducks for cover :-> ).

bastibl:
Yes, this is exactly what I meant. `BlockError` and `Terminate` seem like good terms to me.

Regarding the termination logic in case of a block error, we could consider this block as already done and just wait until the number of `BlockDone` plus `BlockError` corresponds to the number of blocks. Or, as it is currently implemented, decrement `active_blocks` in either case.

I'm not a big fan of the `tick`. It's also what GNU Radio is doing. What happened is that several blocks have broken designs and rely on the tick to proceed. Sounds like non-trivial logic to hide block errors.

PR #83, opened 2022-10-16
pmt: add f64 support for futuresdr

nrot:
Where the Float trait from num-traits can may added. In some places copied the trait for f64. In other case the T could catch Complex<T>. Fixed docs, explicit type required. Test checked. 
_Sorry for my English._


bastibl:
It was cherry-picked on my working state and merged. Unfortunately, GitHub doesn't recognized, but it's upstream as-is.

PR #88, opened 2022-10-23
feat(blocks): SignalSource

loic-fejoz:
Add a new block called `SignalSource` which is the counter part of the GNURadio block.

In particular, it is based on phase being encoded as fixed float and use the same underlying mechanism. It thus provides sine, cosine, square, and triangle signals on `f32`, and wave for `Complex32`. One can also use it with a generic phase to amplitude convertor function.

Moreover, it add an example in order to compare several oscillators implementation. One can retrieve via ZeroMQ the signal in GNURadio to compare the FFT and the stability of the signal.

I also added a `min_item` to the ZeroMQ PubSink in order to avoid sending too many packets.

NB: I would recommend to remove the Oscillator block in favour of this implementation.


bastibl:
Wow, you went deep into the rabbit hole :-) I can do some follow-up commits to remove the oscillator, as you suggested. I'm just not sure about the difference between `offset` and `initial_phase`. Shouldn't only one be required? Or maybe I'm misunderstanding things.

I'd also change the builder to make sample rate and frequency a required parameter, something like:
```rust
SignalSourceBuilder::sin(freq, samp_rate).build();
SignalSourceBuilder::cos(freq, samp_rate).amplitude(2.0).build()
```
Would this be fine for you?

loic-fejoz:
Yes indeed the hole was deep ;-)

 `offset` is about the DC component of the resulting signal, while `initial_phase` is ...well the initial offset in the phase. 
So indeed, we need both.

```
            let a = (self.phase_to_amplitude)(self.nco.phase);
            let a = a * self.amplitude;
            let a = a + self.offset;
```

Otherwise, I am fine with your proposal.

bastibl:
It's merged with the suggested follow-up changes. Thanks!

PR #90, opened 2022-11-06
CW blocks

ratzrattillo:
Hi @bastibl ,

i created some blocks to encode and decode Morse Code messages.

You can try them with the following example program:

```rust
use futuresdr::runtime::{Flowgraph, Runtime};
use futuresdr::blocks::{ConsoleSink, SoapySinkBuilder, SoapySourceBuilder, VectorSource};
use futuresdr::anyhow::{Result};
use futuresdr::blocks::cw::{CharToCWBuilder,CWToCharBuilder, CWToIQBuilder, IQToCWBuilder};
use futuresdr::runtime::buffer::slab::Slab;


#[tokio::main]
async fn main() -> Result<()> {

    let frequency = 1210e6;
    let sample_rate = 250e3;
    let tx_gain = 0.;
    let tx_dev_filter = "driver=bladerf"; //,xb200=auto

    /*let rx_gain = 49.6;
    let rx_dev_filter = "driver=rtlsdr";*/

    let chars = vec!['S', 'O', 'S', ' ', 'S', 'O', 'S'];

    let vector_src = fg.add_block(VectorSource::new(chars));
    let char_to_cw = fg.add_block(CharToCWBuilder::new().build());
    let cw_to_iq = fg.add_block(CWToIQBuilder::new().dot_length(100.).sample_rate(250e3).smoothness(60.).build());
    let soapy_snk = fg.add_block(SoapySinkBuilder::new().freq(frequency).sample_rate(sample_rate).gain(tx_gain).filter(tx_dev_filter).build());

    /*let soapy_src = fg.add_block(SoapySourceBuilder::new().freq(frequency).sample_rate(sample_rate).gain(rx_gain).filter(rx_dev_filter).build());
    let iq_to_cw = fg.add_block(IQToCWBuilder::new().dot_length(100.).sample_rate(250e3).accuracy(70).build());
    let cw_to_char = fg.add_block(CWToCharBuilder::new().build());
    let console_snk = fg.add_block(ConsoleSink::<char>::new(""));*/


    fg.connect_stream(vector_src, "out", char_to_cw, "in")?;
    fg.connect_stream(char_to_cw, "out", cw_to_iq, "in")?;
    //fg.connect_stream(cw_to_iq, "out", soapy_snk, "in")?;
    fg.connect_stream_with_type(cw_to_iq, "out", soapy_snk, "in", Slab::with_size(4096))?;

    /*fg.connect_stream(soapy_src, "out", iq_to_cw, "in")?;
    fg.connect_stream(iq_to_cw, "out", cw_to_char, "in")?;
    fg.connect_stream(cw_to_char, "out", console_snk, "in")?;*/

    let rt = Runtime::new();
    rt.run(fg)?;

    Ok(())
}
```

Use the following Cargo.toml:

```toml
[package]
name = "morse_chat"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
futuresdr = { git = "https://github.com/ratzrattillo/FutureSDR", branch = "cw", features = [ "soapy" ]}
soapysdr = { version = "*" }
tokio = { version = "*", features = ["full"] }
```

If you would like to add it to the runtime i would be happy.
Please feel free to change the code as needed.


bastibl:
Sorry, this was my fault. I had a typo in the repo URL; I just pushed the changes -.-

It moves the blocks to the CW example and extends the existing binary that produces an audio CW signal. The generation of the CW signal was changed to use the existing `ApplyIntoIter` block. Also the smoothing should imho be done with some low pass filter. I extended the audio TX example with these changes. I started an SDR TX example, but this is untested and still lacks the low pass filter.

What's missing is example for the RX part. Would be great to have the corresponding audio and SDR RX examples. Regarding the structure, I'd suggest to factor out the automatic leveling logic into a separate AGC (automatic gain control) block, i.e., SDR -> AGC -> Low Pass? -> Threshold? -> CW Demod.

ratzrattillo:
@bastibl i cannot see the merge of the AGC block yet. However, if that one is done, maybe you might want to check out this CW block examples. Unfortunately my testing capabilities with real hardware are limited until mid january.

bastibl:
The AGC blocks are merged. See here: https://github.com/FutureSDR/FutureSDR/blob/main/src/blocks/agc.rs or the history of the main branch: https://github.com/FutureSDR/FutureSDR/commits/main

Does the audio loopback already work for you?

ratzrattillo:
Ah nice, i did not see that.
The rx-audio example works, but half speed of the tx-audio example. I do not know why however. Nonetheless, feeding IQ-samples to the rx-audio example gives correct audio output.

bastibl:
Thanks for the changes. Following the discussion here: https://github.com/FutureSDR/FutureSDR/issues/104, I'm not sure how to proceed. This is a rather complex example. Maybe it should be it's own project or application; and examples would be more minimal?

I think this PR would require some more changes (e.g., I just tried the `rx_audio` binary, but it panics due to `clap` options. Also, I would expect that `rx_audio` and `tx_audio` interface the audio source/sink and not the SDR, but that doesn't seem to be the case. ). I don't want to annoy you with more requests for change and I don't have the time at the moment to do it myself, since the change set is quite big and would have to be tested w/ hardware etc.

What do you think about putting this in a separate project?

ratzrattillo:
Let me answer the questions regarding this concrete example first:
- The rx-audio example is meant to make morse code symbols received from an SDR device "hearable"
- The tx-audio example is meant to convert text to "hearable" morse code symbols
- The rx-sdr example is meant to decode morse code received via an SDR device to actual text
- The tx-sdr example encodes text to morse code symbols and transmits them via an SDR device

Maybe the naming should be changed in order to avoid this confusion.

I think all examples above are of importance, if we want to include CW examples at all. However this is open to discussion. If we decide to make a seperate project out of the CW decoder program, i would nonetheless like to have it closely connected to FutureSDR. FutureSDR will only become more popular and well known, if more useful tools and examples develop around it.

The goal of many initiators of OpenSource projects is to make themselves obsolete. This is a nice goal, because it means, that everything is documented in a way, that newcomers can contribute with minimal hurdles. FutureSDR however has not yet reached this level. Thus, @bastibl your input is at least for me of great importance. I have no background in Signal Processing and your comments and suggestions help to improve FutureSDR and myself. So please do not hesitate with requests for change.

Regarding your question of having seperate projects instead of examples: Where do you see FutureSDR in version 1.0 compared to 0.27 where we are now? I think this is an essential decision we could discuss in Discord also together with other contributors. :)

bastibl:
A lot of projects have a list of projects that use the library/software. I think, we should start this and put it at a prominent position, for example, the readme.

Maintaining more and more code as part of the main project is not sustainable. Having it as a separate project puts it in a different maintainer/administration domain. Would this be fine for you?

ratzrattillo:
Fine for me!

Issue #85, opened 2022-10-17
Soapy: multiple channel configuration


etschneider:
I've made a first attempt at:
https://github.com/etschneider/FutureSDR/tree/ets_new_soapy

[EDIT: I pushed up some major changes and am adjusting this comment to match. I hope no one was looking at the original code.]

If anyone interested, please review and comment. This isn't pull ready,I just want to see if this general approach is acceptable before spending more time on it.

I made a new `blocks::soapy` sub-module, since the code is more complex. I'm using the `examples/soapy_multichan.rs` as a test case. (Perhaps some `#[ignore]` tests in the  soapy module would be better? Or both?) [EDIT: I've added some integration tests]

A few points:
- The new 'cmd' port is used for all the new configuration methods, and supports the dict format of the GNU Radio Soapy block, among others.
- The `SoapyConfig` type is meant to allow arbitrarily configurations of all channels/directions with a single message
- There is also a new `SoapyCommand` enum, but it doesn't really add anything currently.
- Since the configuration logic is the same for `SoapySource` and `SoapySink`, I've made a generic `SoapyDevice<stream type>` parent for them both. Trying to keep it DRY.
- I tend to think that the old `freq` and `sample_rate` ports should be removed. This would be a breaking change, so???

etschneider:
I'm not very happy with the current config structure, basically a struct full of options for every setting:

```rust
pub struct SoapyConfig {
    pub chan: Option<usize>,
    pub dir: SoapyConfigDir,
    pub sample_rate: Option<f64>,
    pub freq: Option<f64>,
    pub gain: Option<f64>,
    ...
```
Seems pretty wasteful, especially assuming that the configuration options will grow to support more of the Soapy device API.

I'll play around with something like:

```rust
enum SoapyConfigItem {
    Gain(f64),    
    Freq(f64),
    ...

struct SoapyConfig(Vec<SoapyConfigItem>);
```

The challenge is dealing with the channel number and direction options in a flexible and elegant way. Perhaps just a chan/dir enum item that applies to all subsequent items in the vec.

bastibl:
The message handler for the dict is very cool! Maybe it's the way to go and we should remove the other handlers. I had a look at the Soapy API. It's pretty big. I think we should have a rough plan about the way we want to proceed. At the moment, we add stuff as needed, but I don't have a good idea about a more complete implementation. Do you have any thoughts? Should we support the API calls through the dict?
[...]

PR #91, opened 2022-11-08
soapy: multi-channel configuration.

etschneider:
Supports individual channel configuration via builder and runtime msg ports.
Fixes: #85


bastibl:
thanks

PR #95, opened 2022-11-23
unsafe: rm unneeded Send/Sync impls

etschneider:
Figured it is better to let the compiler infer the traits when possible rather than forcing/faking it.


bastibl:
Cool, thanks!

PR #98, opened 2022-11-29
Initial Commit AGC Block

ratzrattillo:
Initial version of an AGC block.
However i still seem to have issues with the Pmt::Any part at e.g. https://github.com/FutureSDR/FutureSDR/compare/main...ratzrattillo:FutureSDR:sw_agc?expand=1#diff-a72e9fc325a0c7f1e0db0691e8314312b3724262a5fd8c0b997e8ce247791fefR70-R80

Setting the values is not yet working due to unknown Type id. Have to check that another day.
If you know a quick fix please let me know!

### DIFF
--- a/src/blocks/agc.rs
+++ b/src/blocks/agc.rs
@@ -1,5 +1,7 @@
+use num_complex::ComplexFloat;
 use futuresdr_pmt::Pmt;
 use futures::FutureExt;
+use rustfft::num_traits::ToPrimitive;
 
 use crate::anyhow::Result;
 use crate::runtime::Block;
@@ -13,7 +15,7 @@ use crate::runtime::StreamIoBuilder;
 use crate::runtime::WorkIo;
 
 
-pub struct AGC {
+pub struct AGC<T> {
     // Minimum value that has to be reached in order for AGC to start adjusting gain.
     squelch: f32,
     // maximum gain value (0 for unlimited).
@@ -26,9 +28,12 @@ pub struct AGC {
     adjustment_rate: f32,
     // (Boolean) Set when gain should not be adjusted anymore, but rather be locked to the current value
     gain_lock: u32,
+    _type: std::marker::PhantomData<T>,
 }
 
-impl AGC
+impl<T> AGC<T>
+    where
+        T: Send + Sync + ComplexFloat + 'static,
 {
     pub fn new(
         squelch: f32,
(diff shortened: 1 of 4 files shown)


src/blocks/agc.rs
```diff
@@ -0,0 +1,257 @@
...
+use crate::runtime::MessageIo;
+use crate::runtime::MessageIoBuilder;
+use crate::runtime::StreamIo;
+use crate::runtime::StreamIoBuilder;
+use crate::runtime::WorkIo;
+
+
+pub struct AGC<T> {
+    // Minimum value that has to be reached in order for AGC to start adjusting gain.
+    squelch: T,
```
etschneider, line 19: Why use type T for members? I think it would make more sense to just use a fixed type, like f32. All these gain related values should be real, so if I try to use a complex sample type it would be weird. If it is desired to use different real gain types (f32, f64) then an additional type could be used to differentiate it from the sample type. I'd just stick with f32 for now.

src/blocks/agc.rs
```diff
@@ -0,0 +1,257 @@
...
+                               }.boxed()
+                           },
+                )
+                .add_input("max_gain",
+                           |block: &mut AGC<T>,
+                            _mio: &mut MessageIo<AGC<T>>,
+                            _meta: &mut BlockMeta,
+                            p: Pmt| {
+                               async move {
+                                   if let Pmt::Any(ref r) = &p {
```
etschneider, line 70: If member types are just f32 then this would be much easier with just Pmt::F32(), no Pmt::Any complications.

src/blocks/agc.rs
```diff
@@ -0,0 +1,257 @@
...
+pub struct AGC<T> {
+    // Minimum value that has to be reached in order for AGC to start adjusting gain.
+    squelch: T,
+    // maximum gain value (0 for unlimited).
+    max_gain: T,
+    // initial gain value.
+    gain: T,
+    // reference value to adjust signal power to.
+    reference_power: T,
+    // the update rate of the loop.
```
etschneider, line 26: The meaning of "update_rate" was unclear to me by name before looking at the code. I first thought you might be doing periodic amplitude sampling. Perhaps something like "slew_rate", "gain_rate", or "adjustment_rate"? Pretty minor though, ignore at will.

Also, might as well use "doc comments" "///" here.

etschneider (commented):
I put a couple comments inline. I think avoiding generic type T for gain values makes sense here. That also avoids the `Pmt::Any` downcasting issues.

src/blocks/agc.rs
etschneider: GNU Radio just uses the term "rate". 

https://wiki.gnuradio.org/index.php/AGC

etschneider:
@ratzrattillo, I think there was a misunderstanding. The sample type should still be generic <T>, only the inner struct types as fixed real types, like f32.

E.g.
```rust
pub struct AGC<T> {
    // Minimum value that has to be reached in order for AGC to start adjusting gain.
    squelch: f32,
    // maximum gain value (0 for unlimited).
    max_gain: f32,
    // initial gain value.
    gain: f32,
    // reference value to adjust signal power to.
    reference_power: f32,
    // the update rate of the loop.
    adjustment_rate: f32,
    // (Boolean) Set when gain should not be adjusted anymore, but rather be locked to the current value
    gain_lock: u32,
    _type: PhanomData<T>,
}
```

Then something like:
```rust
impl<T> AGC<T>
{
    fn scale(&mut self, input: T) -> T {...}
}
```

```rust
impl<T: Send + 'static> Kernel for AGC<T> {
...
}
```

Where type T is the sample type of the input and output buffers. The sample type could be `f32`, `Complex32`, or whatever.

bastibl:
I rebased and merged it with some follow-up commits. Thank you both!

PR #100, opened 2022-12-04
Fix tag debug

AndersKaloer:
The current `TagDebug` block violates the following assert when the input type is different from `u8`: https://github.com/FutureSDR/FutureSDR/blob/main/src/runtime/stream_io.rs#L79.

I believe the implementation is supposed to be like the one implemented in this pull request.

I also created a simple `TagDebug` example, but I did not include it in the pull request (https://github.com/AndersKaloer/FutureSDR/tree/tag-debug-example). The example fails with the current `TagDebug` implementation, but works with this fix.


bastibl:
Great, thank you! I missed this when trying to make the ports more type-safe. I created `examples/debug` for code examples that cannot easily be made into unit tests. You could add the tag example there if you wanted.

PR #101, opened 2022-12-06
Add simple example with TagDebug

AndersKaloer:
As suggested in #100, this adds a simple example with the TagDebug block to `examples/debug`.


bastibl:
thanks!

PR #102, opened 2022-12-23
ADS-B decoder example

AndersKaloer:
I implemented an ADS-B decoder with FutureSDR, and thought it could be nice to include as an example.

It also contains a simple web-based map that uses the ctrl_port to interact with the flowgraph.

It relies on [adsb_deku](https://github.com/rsadsb/adsb_deku) for decoding the packets. While it supports a lot of formats, I have only implemented the most common ones, and the ones that were easiest to integrate.

I have not performed any actual comparison to other ADS-B implementations, but it seems to perform quite well on my recordings and with an RTL-SDR with the stock telescopic antenna.

Considering that many people are likely to use this example (along with the others) as an example of how to implement a decoder with FutureSDR, I think it should be as close as possible to "idiomatic" FutureSDR. So suggestions are very welcome. For instance, maybe it could make sense to merge some of the blocks, such as the preamble detector and demodulator.

Also note that I copied the web server code from the zigbee example and only made some minor modifications. I am not familiar with gulp, so there might be some unnecessary stuff that can be removed.

### DIFF
--- a/examples/adsb/Cargo.toml
+++ b/examples/adsb/Cargo.toml
@@ -18,4 +18,4 @@ serde = { version = "1.0", features = ["derive"] }
 serde_with = "2.1.0"
 serde_json = "1.0.89"
 env_logger = "0.10.0"
-adsb_deku = { git = "https://github.com/rsadsb/adsb_deku" }
+adsb_deku = "0.6.2"
--- a/examples/adsb/Cargo.toml
+++ b/examples/adsb/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "adsb_demod"
+version = "0.1.0"
+edition = "2021"
+
+[workspace]
+
+[[bin]]
+name = "listen-adsb"
+path = "src/bin/listen_adsb.rs"
+
+[dependencies]
+clap = { version = "4.0.27", features = ["derive"] }
+futuresdr = { path = "../..", features=["soapy", "audio"] }
+futuredsp = { path = "../../futuredsp" }
+futuresdr-pmt = { path = "../../pmt" }
+serde = { version = "1.0", features = ["derive"] }
+serde_with = "2.1.0"
+serde_json = "1.0.89"
+env_logger = "0.10.0"
+adsb_deku = { git = "https://github.com/rsadsb/adsb_deku" }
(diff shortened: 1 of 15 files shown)


wcampbell0x2a:
I just so happen to follow this project so I saw this pull request. I'm the author of adsb-deku and this is sweet!

examples/adsb/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+[dependencies]
+clap = { version = "4.0.27", features = ["derive"] }
+futuresdr = { path = "../..", features=["soapy", "audio"] }
+futuredsp = { path = "../../futuredsp" }
+futuresdr-pmt = { path = "../../pmt" }
+serde = { version = "1.0", features = ["derive"] }
+serde_with = "2.1.0"
+serde_json = "1.0.89"
+env_logger = "0.10.0"
+adsb_deku = { git = "https://github.com/rsadsb/adsb_deku" }
```
wcampbell0x2a, line 21: I pushed an updated library that builds, should be able to change this to `adsb_deku = "0.6.2"`.

Tested this myself and it worked fine.

AndersKaloer: Thanks! It works for me as well.

examples/adsb/src/bin/listen_adsb.rs
```diff
@@ -0,0 +1,142 @@
...
+use std::time::Duration;
+
+#[derive(Parser, Debug)]
+#[clap(version)]
+struct Args {
+    /// Antenna
+    #[clap(short, long)]
+    antenna: Option<String>,
+    /// Soapy Filter
+    #[clap(short, long)]
```
hagmonk, line 27: Suggest removing `short` from `filter` here as it clashes with `file` later on and causes a panic:

```
thread 'main' panicked at 'Command adsb_demod: Short option names must be unique for each argument, but '-f' is in use by both 'filter' and 'file'',
```

I think it's more UX friendly to expect `-f` to be short for `--file` which is why I propose removing short here, but, more important to get rid of the panic 😬

AndersKaloer: Ah, I did not catch that. Thanks!
At the same time, I updated to use clap 4 syntax.

bastibl:
Thanks! This is absolutely awesome :-) I really like how FutureSDR is used here (i.e., it does the stream processing, uses a 3rd party crate for decoding, and exposes data through a web interface). That's exactly how I hoped it would be used. Makes a great example.

PR #103, opened 2022-12-23
unsafe: copy_nonoverlapping -> copy_from_slice

etschneider:
Use types in IO slice_unchecked to avoid manual type size computation. Also add Copy trait bound to block type.

### DIFF
--- a/examples/wlan/src/delay.rs
+++ b/examples/wlan/src/delay.rs
@@ -17,12 +17,12 @@ enum State {
     Skip(usize),
 }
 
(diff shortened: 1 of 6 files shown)


etschneider:
I've been sitting on this for a while and didn't want it to get too stale.

I checked that the actual copies compile to identical code (memcpy) so no performance loss.

The use of typed slices seems better. Hopefully I'm not missing something. There are *a lot* more blocks that could be modified if it is agreed that this is a better style.

bastibl:
Great, thanks!

PR #109, opened 2023-03-05
Add documentation for examples

JuliDi:
Since there are already many different cool examples, it makes quite difficult for new users to figure out what they all do.
Not all examples have documentation in the code and some require extra setup (i.e. going to the folder and executing `cargo run` does not work) or have some other constraints.

This PR should serve as a starting point for giving an overview of the examples and adding some "how-to" for each example.

It's probably best if the authors of the examples add some info themselves. I am only somewhat familiar with the fm-receiver, so I took that as a start.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,8 @@
+# Examples
+
+## FM Radio
+`examples/fm-receiver`
+
+A simple FM receiver that you can tune to nearby radio stations.
+
+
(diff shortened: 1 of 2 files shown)


bastibl:
Great, thanks for starting that! (I was also working on API documentation recently. At least, I wanted to have *something* for all public items and improve over time.)
Regarding the draft status: We could just merge this PR and track the example documentation in an issue?

PR #110, opened 2023-04-01
fix(audio): Channels issue on MS/Windows

loic-fejoz:
Solve channels number issue when, for instance you request 1 channel but stream config requires at least 2.

### DIFF
--- a/src/blocks/audio/audio_sink.rs
+++ b/src/blocks/audio/audio_sink.rs
@@ -70,6 +70,7 @@ impl AudioSink {
             if let Ok(configs) = d.supported_output_configs() {
                 let mut v = Vec::new();
                 for c in configs {
+                    // println!("{:?}", c);
                     let min = c.min_sample_rate().0;
                     let max = c.max_sample_rate().0;
                     if min >= 10000 {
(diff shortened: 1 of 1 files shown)


bastibl:
I think, we shouldn't try some magic, when the user parameters are bogus. It might be easier to extend the interface to the cpal configs, for example, add a function to get all supported configs: https://github.com/FutureSDR/FutureSDR/blob/main/src/blocks/audio/audio_sink.rs#L70
And we could think about changing the constructor to accept a cpal config. We anyway need to extend this API. For example to select the device, etc. So adding more and more magic for wrong parameters might cause complexity later.

loic-fejoz:
Good point from FutureSDR perspective where indeed we expect one to adjust perfectly for is own hardware.

From `fsdr-cli` perspective, I need to support those config's parameters from GNU Radio flowgraph for instance:
```yaml
- name: audio_sink_0
  id: audio_sink
  parameters:
    affinity: ''
    alias: ''
    comment: ''
    device_name: ''
    num_inputs: '1'
    ok_to_block: 'True'
    samp_rate: '48000'
```

Note that the config is not bogus, we are really generating mono audio while cpal absolutely requires a stereo flow in my MS/Windows setup.

Let's reject this pull request.

PR #112, opened 2023-04-24
Update dependencies

ratzrattillo:
Update dependencies

Adjust tower error handling to v0.4.0 standards.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -117,7 +117,7 @@ async-native-tls = { version = "0.5", optional = true }
 async-net = "1.7.0"
 async-task = "4.4.0"
 async-tungstenite = "0.21.0"
-axum = "0.6.17"
+axum = "0.6.18"
 blocking = "1.3"
 concurrent-queue = "2.2"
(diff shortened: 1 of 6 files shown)


ratzrattillo:
Just saw that some unit tests failed. I will check and try to fix. When commiting i just ran "cargo test" on Windows.

ratzrattillo:
@bastibl would you mind fixing the Vulkan issues? I do not have devices to test.

Another option would be to not update the vulkano crate.

bastibl:
Thanks for the PR. I was looking into the Vulkan changes but they seem a bit more involved. Ideally, the code should also be a bit more typed, i.e., typed Vulkan CPU buffers. But that's a non trivial change. Unfortunately, I won't have the time to look into it in the next few days. So you could just leave out the Vulkan changes for now.

bastibl:
I just pushed a fixed for the clippy warning. (Let's see, if more pop up during the next CI run.)
Are you otherwise done with  the PR?

PR #113, opened 2023-05-05
Add autolock

ratzrattillo:
Fix AGC property assignment bug

### DIFF
--- a/src/blocks/agc.rs
+++ b/src/blocks/agc.rs
@@ -27,6 +27,8 @@ pub struct Agc<T> {
     adjustment_rate: f32,
     /// Set when gain should not be adjusted anymore, but rather be locked to the current value
     gain_locked: bool,
+    /// Set when gain should be automatically locked, when the reference power is reached for the first time
+    autolock: bool,
     _type: std::marker::PhantomData<T>,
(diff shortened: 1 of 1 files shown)


bastibl:
Is this a bug fix or a new feature?

ratzrattillo:
Its both, sorry for not separating it properly.
It contains the autlock feature, which might make sense, if you want to increase the gain constantly, until you hit the reference power. From this point on, gain will be locked.

The bugfix is a previously wrong assignment of the adjustment_rate property to self.sqelch instead of self.adjustment_rate.

bastibl:
The PR also seems to update an unrelated dependency. How about we merge the dependency update and the bug fix. And the block is moved to `fs-blocks` together with the example? Then you can extend the block there as needed (e.g., the auto-lock feature). I don't know much about the block and it would take me very long to thoroughly review PRs.

ratzrattillo:
Sounds like an idea. Then i suggest you discard this pull request and update the dependency by yourself when doing the next release (Should not have breaking changes).
Please leave the AGC block still in the FutureSDR main repository, until i ported it to fsdr-blocks repo.

PR #115, opened 2023-06-01
fix(futuredsp): FIR assertions and types

loic-fejoz:
Fix assertion for cutoff of filter.
Add more generic implementation of taps so as to handle Complex32 IQ stream with Complex32 taps.


bastibl:
Great, thank you!

PR #135, opened 2024-03-23
Fix wasm example

### LINKED ISSUE #133: wasm example fails to build
Using the current main branch, the wasm example fails to build. This also happens when I try to compile a wasm project with FutureSDR as a dependency. I will try to figure out what's the issue and report back. Any hints on where to look are welcome!

```
error[E0433]: failed to resolve: could not find `UsbControlTransferParameters` in `web_sys`
   --> /home/user/Git/FutureSDR_main/src/blocks/wasm/hack_rf.rs:249:34
    |
249 |         let parameter = web_sys::UsbControlTransferParameters::new(
    |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ could not find `UsbControlTransferParameters` in `web_sys`
```

This indicates that web-sys is somehow not working, maybe a missing feature.

JuliDi:
Close #133 by adding compiler flags

### DIFF
--- a/examples/wasm/.cargo/config.toml
+++ b/examples/wasm/.cargo/config.toml
@@ -1,0 +1,2 @@
+[build]
+rustflags = ["--cfg=web_sys_unstable_apis", "-C", "target-cpu=generic"]
--- a/examples/wasm/Makefile
+++ b/examples/wasm/Makefile
@@ -1,6 +1,6 @@
 .PHONY: build
 
 build:
-	RUSTFLAGS='-C target-cpu=generic' wasm-pack build --target web --out-name wasm --out-dir ./dist --dev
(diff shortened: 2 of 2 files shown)


bastibl:
Thank you! There are also the zigbee and wgpu examples, which are actually better examples. I'll try to overhaul this example as well, getting rid of the node dependency and Makefile.

JuliDi:
Thank you, I will check these out, as well.
Is there any workaround for the missing throttle block on wasm? My impression is that the wasm examples either terminate after N samples or use an actual SDR with a limited data rate.
(Opened a discussion topic to gather some ideas for porting it here, for reference: https://github.com/FutureSDR/FutureSDR/discussions/134)

PR #136, opened 2024-03-24
Fix use of `std::time::Instant` in Throttle block

JuliDi:
In wasm you cannot use `std::time::Instant` so their uses have to be replaced. I hopefully caught all cases where it might be used from a wasm context.


bastibl:
thanks!

JuliDi:
I think this should be working now, hope I didn't break anything during rebase.

PR #139, opened 2024-05-26
Implement GUI blocks and default interface

KoffeinFlummi:
Adds a new sub-module of GUI blocks (message sliders, spectrum and waterfall plots for now), gated behind the `gui` feature. These blocks are present in the flowgraph and communicate with some sort of handle which is drawn by the frontend implementation. Applications can either draw these handles themselves or use a "batteries-included" UI. I have implemented frontends using `egui`, and a basic TUI interface using `textplots`. The frontend is selected with the appropriate feature flag.

GUI block builders implement `build_detached()` and `build()`. The latter behaves just like other block builders. Behind the scenes the GUI handle is attached to the block, and then extracted when passing the flowgraph to the default UI. `build_detached` returns both the block and the GUI handle, which can then be drawn manually.

The `gui` examples has binaries for both the batteries-included UI and a custom one:
`egui` example screencast
`textplots` example screencast

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,10 @@ wgpu = ["dep:wgpu"]
 zeromq = ["dep:zmq"]
 zynq = ["dep:xilinx-dma"]
 
+gui = []
+egui = ["gui", "dep:egui", "dep:eframe", "dep:egui_plot", "dep:colorgrad"]
+textplots = ["gui", "dep:textplots", "dep:drawille", "dep:console", "dep:ctrlc", "dep:rgb"]
+
 [[bench]]
 name = "flowgraph"
 harness = false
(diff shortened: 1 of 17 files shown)


bastibl:
Hi, I'm so sorry for the late reply. I took me forever to look into this.

As I already wrote on Discord, I really love the examples and the text plots in particular. Also the fact that this can still be rebased, shows that it is not invasive to the code base.

What I mainly wonder is, whether the GUIs should become part of the runtime, or if we should try to keep them separate. In the other egui example, I terminate the FutureSDR flowgraph in a ChannelSink and the GUI does the plotting outside FutureSDR:

https://github.com/FutureSDR/FutureSDR/blob/main/examples/egui/src/bin/combined.rs#L82

I think the approach of the blocks in this PR is very similar, just that the channel and the GUI are encapsulated in a FutureSDR block.

Do you see an advantage of having the GUI as part of FutureSDR? Or maybe we move the GUI widgets to a separate futuredsr-egui crate that provides widgets and helper functions, similar to `Gui::run(fg)`.

metasim:
It would be great to have this content in a (separate sub-crate)! Nice work!

PR #140, opened 2024-06-30
Telemetry

ratzrattillo:
Add the possibility to collect telemetry data on a FutureSDR Application


bastibl:
I'll close this for now, since its stale for quite some time and no longer applies to main. Maybe we can have a fresh look at this in the future.  It would be nice to have to observe runtime metrics while the flowgraph is running.

ratzrattillo:
Good idea to close this. i had a different understanding of telemetry when i started this. This should be redone in a different way.

PR #141, opened 2024-07-05
Add FFT Size MessageIO to FFT Block

K4HVH:
Adding an `fft_size` MessageIO block to the FFT block. This allows for the FFT size to be changed during runtime.


bastibl:
Cool, thank you!

PR #142, opened 2024-07-19
Add autolock to AGC block

ratzrattillo:
Add autolock to AGC block

Allow logarithmic raise of power to desired power level
Configurable adjustment speed of power level

### DIFF
--- a/src/blocks/agc.rs
+++ b/src/blocks/agc.rs
@@ -1,6 +1,3 @@
-use num_complex::ComplexFloat;
-use rustfft::num_traits::ToPrimitive;
-
 use crate::anyhow::Result;
 use crate::runtime::Block;
 use crate::runtime::BlockMeta;
@@ -12,6 +9,7 @@ use crate::runtime::Pmt;
 use crate::runtime::StreamIo;
 use crate::runtime::StreamIoBuilder;
 use crate::runtime::WorkIo;
+use num_complex::ComplexFloat;
 
 /// Automatic Gain Control Block
 pub struct Agc<T> {
@@ -186,9 +184,9 @@ where
         // Gain is only exactly 1.0 when the AGC block is freshly initialized
         // We then can set the gain to a suitable value to scale
         // the input power to e.g. 99% of the reference power immediately.
-        if self.gain == 1.0 {
-            self.gain = (self.reference_power / input.abs().to_f32().unwrap()) * 0.99;
-        }
+        // if self.gain == 1.0 {
+        //     self.gain = (self.reference_power / input.abs().to_f32().unwrap()) * 0.99;
+        // }
 
         // The gain adjustment rate should not be fixed, but depending
         // on the input power in relation to the reference power (Thus actually the gain).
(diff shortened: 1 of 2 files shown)


src/blocks/agc.rs
```diff
@@ -157,10 +180,51 @@ where
...
-            self.gain +=
-                (self.reference_power - output.abs().to_f32().unwrap()) * self.adjustment_rate;
-            self.gain = self.gain.min(self.max_gain);
+        // if the input power is very low or very high compared to the reference power, we still want to reach the reference power quickly.
+        // Thus we should make the gain also dependent on the input_power in relation to the reference power.
+
+        // Gain is only exactly 1.0 when the AGC block is freshly initialized
+        // We then can set the gain to a suitable value to scale
+        // the input power to e.g. 99% of the reference power immediately.
+        if self.gain == 1.0 {
```
bastibl, line 189: I would avoid such tests. You could maybe add a flag, but I think such a mechanism could also lead to problems. Consider a sine with perfect power level, which also is zero (or close to zero) at the first sample. So this would make it worse. So I would consider to delete this initial adjustment.

src/blocks/agc.rs
```diff
@@ -185,10 +249,11 @@ where
         let m = std::cmp::min(i.len(), o.len());
         if m > 0 {
             for (src, dst) in i.iter().zip(o.iter_mut()) {
-                if src.abs().to_f32().unwrap() > self.squelch {
+                let input_power = src.abs().to_f32().unwrap();
+                if input_power > self.squelch {
                     *dst = self.scale(*src);
                 } else {
-                    *dst = T::from(0.0).unwrap();
+                    *dst = T::from(self.reference_power).unwrap();
```
bastibl, line 256: If something is below the squelch, I think it should be zero.

bastibl (commented):
Thanks for your PR! I added two small comments.

Regarding the dynamic adjustments: If this makes sense might depend on the waveform of the signal. It seems to be designed for rather constant signals. Did you adapt this from an existing implementation?

Btw: When looking at the code, I just noticed that the "reference power level" is actually a reference amplitude, no?

ratzrattillo:
> Thanks for your PR! I added two small comments.
> 
> Regarding the dynamic adjustments: If this makes sense might depend on the waveform of the signal. It seems to be designed for rather constant signals. Did you adapt this from an existing implementation?
> 
> Btw: When looking at the code, I just noticed that the "reference power level" is actually a reference amplitude, no?

I did not adapt this from an existing implementation. It is the outcome of my own thoughts. I will fix the part with the power level and also implement the changes suggested regarding the initial adjustment and squelch level. I do not have a good understanding of signal processing, that is why this AGC probably is not working well for all kinds of signals. I used it for sending morse code.

bastibl:
I see. So since this is a custom algorithm that is designed and only tested for your application, I would recommend to put move the block to the application. That's usually what I do for the example applications as well. What do you think?

We already considered it some time ago: https://github.com/FutureSDR/FutureSDR/pull/113#issuecomment-1543602867

Moving the block should be straight forward (copy the file to the other module, add it to `lib.rs`/`mod.rs`, and change the dependencies from `crate::..` to `futuresdr::...`.

ratzrattillo:
Then lets do it that way! I completely forgot we already talked about this in the past!

PR #148, opened 2024-09-30
futuredsp: fix iir memory update

tkj666:
Updating `memory` in ascending order would set all the elements to `memory[0]`. Fix it.


bastibl:
Thank you!

PR #150, opened 2024-10-22
Updated `message_handler` macro to emit fully qualified Result type in return.

### LINKED ISSUE #149: `#[message_handler]` hygiene: macro assumes import of  `futuresdr::anyhow::Result`
This code fails to compile for reasons that are difficult to understand at first:

```rust
use futuresdr::macros::message_handler;
use futuresdr::runtime::{Block, BlockMetaBuilder, BlockMeta, Kernel, MessageIo, MessageIoBuilder, Pmt, StreamIoBuilder, WorkIo};

struct MsgThing;

impl MsgThing {
    fn new() -> Block {
        Block::new(
            BlockMetaBuilder::new("MsgThing").build(),
            StreamIoBuilder::new()
                .build(),
            MessageIoBuilder::new()
                .add_input("in", Self::in_port)
                .build(),
            Self
        )
    }

    #[message_handler]
    async fn in_port(
        &mut self,
        _io: &mut WorkIo,
        _mio: &mut MessageIo<Self>,
        _meta: &mut BlockMeta,
        p: Pmt,
    ) -> futuresdr::anyhow::Result<Pmt> {
        Ok(Pmt::Ok)
    }
}

impl Kernel for MsgThing {}
```

Error message:

```rust
error[E0107]: enum takes 2 generic arguments but 1 generic argument was supplied
   --> src/main.rs:19:5
    |
19  |     #[message_handler]
    |     ^^^^^^^^^^^^^^^^^^
    |     |
    |     expected 2 generic arguments
    |     supplied 1 generic argument
    |
note: enum defined here, with 2 generic parameters: `T`, `E`
   --> /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/core/src/result.rs:527:10
    |
527 | pub enum Result<T, E> {
    |          ^^^^^^ -  -
    = note: this error originates in the attribute macro `message_handler` (in Nightly builds, run with -Z macro-backtrace for more info)
help: add missing generic argument
    |
19  |     #[message_handler], E
    |                       +++

```

If you add `use futuresdr::anyhow::Result` to the current scope it compiles, but this conflicts with the use of crate-local `Result` types, and is confusing due to the return type of `in_port` being fully qualified. 

I suspect the problem is that the macro doesn't fully qualify the `Result` type here:

https://github.com/FutureSDR/FutureSDR/blob/f44b52c5e3c979af239b7706d1572383ec2efe90/crates/macros/src/lib.rs#L552

metasim:
Closes #149

### DIFF
--- a/crates/macros/src/lib.rs
+++ b/crates/macros/src/lib.rs
@@ -549,7 +549,7 @@ pub fn message_handler(
             #mio: &'a mut MessageIo<Self>,
             #meta: &'a mut BlockMeta,
             #pmt: Pmt,
-        ) -> std::pin::Pin<Box<dyn std::future::Future<Output = Result<Pmt>> + Send + 'a>> {
+        ) -> std::pin::Pin<Box<dyn std::future::Future<Output = crate::anyhow::Result<Pmt>> + Send + 'a>> {
             use crate::futures::FutureExt;
             Box::pin(async move {
                 #(#body)*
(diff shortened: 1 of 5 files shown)


bastibl:
This macro is a bit of a disaster :-) I think, it just ignores the return parameter in the code and puts `anyhow::Result<Pmt>` there. I don't know, if there are better ways. This way, at least the code looks, as one would expect.

Independent from that, this PR definitely makes things more specific, which is great. Thanks!

PR #151, opened 2024-10-22
Added tracking of Stream port type information 

metasim:
When an attempt at connecting incompatible ports is made, you now get a more descriptive error.

For example, this code:

```rust
let fft = Fft::new(1024);
let sink = NullSink::<[Complex<f32>; 1024]>::new();
let fft = fg.add_block(fft);
let sink = fg.add_block(sink);
fg.connect_stream(fft, "out", sink, "in")?;
```

Will now issue an error with this message:

```
Connect error, incompatible ports: Fft_0.out: num_complex::Complex<f32> -> NullSink_0.in: [num_complex::Complex<f32>; 1024]
```


src/runtime/mod.rs
```diff
@@ -218,8 +218,16 @@ pub enum Error {
...
+    #[error("Connect error, incompatible ports: {src_block}.{src_port}: {src_type} -> {dst_block}.{dst_port}: {dst_type}")]
+    #[allow(missing_docs)]
+    ConnectError {
+        src_block: String,
+        src_port: String,
+        src_type: String,
+        dst_block: String,
+        dst_port: String,
+        dst_type: String,
+    },
```
metasim, line 230: I could switch this back to a tuple variant, especially since this error is only instantiated in one place, but giving the fields names makes it a bit easier to remember what's expected if it's every instantiated elsewhere.

src/runtime/stream_io.rs
```diff
@@ -30,6 +30,7 @@ pub struct StreamInput {
     name: String,
     item_size: usize,
     type_id: TypeId,
+    pub(crate) type_name: &'static str,
```
metasim, line 33: Could create a `pub(crate)` accessor instead. But this should be crate-private to keep users from expecting this to be a stable identifier (which its not).

bastibl:
Awesome, much better error message. Thank you!

PR #155, opened 2024-10-25
Lifted `Keep1InN` example block into top-level `MovingAvg` block.

metasim:
Includes porting examples to new Block and removing old ones.

In order to get `examples/egui` working on macOS, updated GLSL code to 3.0 syntax.

## Testing

Notes on testing:

* Added stand-alone unit test in `tests/moving_avg.rs`
* `examples/egui/src/bin/combined.rs` Works after GLSL code update.
* Unable to get `examples/egui/src/bin/{fg.rs,egui.rs}` to run due to websocket errors (each throws an error assuming the other is running already, I think).
* Couldn't get `examples/spectrum` to run due to websocket errors.
* Unable/unsure how to test `examples/android-hw`

## Other

* How do we differentiate against `examples/m17/src/moving_average.rs`?


bastibl:
Regarding our discussion on Discord: the issue for me was that the I/O logic of the MovingAverage was a bit off when the downstream block was blocking or very slow. I'll patch in a follow-up. Thank you!

PR #157, opened 2024-10-28
Add CPU pin scheduler

daniestevez:
This adds a new scheduler that takes a HashMap that indicates to which CPU each block should be pinned.

---

Background: To do some benchmarking, I wanted to have a scheduler in which I could specify manually in which CPU each block should run, similarly to using `set_processor_affinity()` in GNU Radio (this is one of the things I want to compare with in my benchmarks). I have implemented this as a variant of the `FlowScheduler` that takes a `HashMap` defining in which worker each block should go. The reason why I've implemented this custom scheduler in the `futuresdr` crate is because it doesn't seem possible to implement schedulers in an external crate (for instance, `topology.blocks` is private). So now that I have this written, I was wondering if it made sense to contribute this upstream. My initial benchmarks show that this custom scheduler is worse than the regular smol scheduler with no CPU pinning. So this custom scheduler doesn't have much practical use. It can be useful for benchmark/comparison and also in case someone wants to investigate why it performs worse, since the kind of flowgraphs I'm benchmarking should be able to benefit from manual CPU pinning.

I've marked the PR as draft because I'm asking for comment about whether it makes sense to merge this to upstream.

### DIFF
--- a/src/runtime/scheduler/cpu_pin.rs
+++ b/src/runtime/scheduler/cpu_pin.rs
@@ -1,0 +1,166 @@
+use super::flow::FlowExecutor;
+use crate::runtime::{config, scheduler::Scheduler, BlockMessage, FlowgraphMessage, Topology};
+use async_io::block_on;
+use async_lock::Barrier;
+use async_task::Task;
+use futures::channel::{
+    mpsc::{channel, Sender},
+    oneshot,
+};
+use futures_lite::future::Future;
+use slab::Slab;
+use std::{collections::HashMap, fmt, sync::Arc, thread};
+
+type CpuPins = HashMap<usize, usize>;
+
+/// CPU pin scheduler
+///
+/// Pins blocks to worker threads fixed to CPUs according to a hashmap.
+#[derive(Clone, Debug)]
+pub struct CpuPinScheduler {
+    inner: Arc<CpuPinSchedulerInner>,
+}
+
+struct CpuPinSchedulerInner {
+    executor: Arc<FlowExecutor>,
+    workers: Vec<(thread::JoinHandle<()>, oneshot::Sender<()>)>,
+    cpu_pins: CpuPins,
+}
+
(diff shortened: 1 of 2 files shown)


daniestevez:
Great. I'm marking this PR as ready for review. Happy to make any changes to make this more useful for your use cases or for a wider context.

I agree that this is much different from what GNU Radio 3.10 does, but it is the closest I can get to a fair comparison given that GNU Radio 3.10's scheduler is inflexible (I guess I could also benchmark the FutureSDR thread-per-block scheduler to have a closer comparison; maybe I'll try).

The kind of flowgraph I'm benchmarking is something like
```
Null Source -> Math block -> Math block -> ... -> Math block -> Benchmark Sink
```
Here `Null Source` is really a "zero-cost" null source (It doesn't memset the output buffer to zero. It just "pretends" that it produces samples, but doesn't ever write to the buffer), Math block is a very optimized implementation of `y[n] = a * x[n] + b` (with everything `f32`), and `Benchmark Sink` just counts samples and measures elapsed time every once in a while.

I'm doing several variations of this flowgraph, but one important example is pinning each `Math block` to a different CPU, and making the `Null Source` and `Benchmark Sink` share the CPU with the `Math block` they're directly connected to. In this case there is basically a single block per CPU (`Null Source` and `Benchmark Sink` are supposed to be almost zero-cost). So in principle there should be little difference in FutureSDR between thread-per-block with each thread pinned to the corresponding CPU, and this CPU pin scheduler (the difference is whether `Null Source` and the downstream `Math block` share a thread or not, and likewise for `Benchmark Sink`).

In GNU Radio 3.10 I'm finding that there is a lot of overhead in calling these blocks. `Null Source` and `Benchmark Sink` are far from being zero-cost, and use almost the same CPU as each `Math block`.

I'm also doing a comparison with GNU Radio 4.0. There I can have a custom scheduler that does more or less the same as this CPU pin scheduler.

At some point I'll be publishing code and results for all these benchmarks.

PR #161, opened 2024-10-31
Added additional `From` & `TryFrom` converters for `Pmt`.

metasim:
* Per [doc.rust-lang.org](https://doc.rust-lang.org/std/convert/trait.TryInto.html), libraries should implement `TryFrom` instead of `TryInto`, as generic implementations of `TryInto` already exist when `TryFrom` exists. Therefore, existing implementations were converted.
* For each `TryFrom<Pmt> for T` where `T: Copy`, added `TryFrom<&Pmt> for T` to for better ergonomics with reference bindings where implicit copy isn't a concern.
* Added `From<T> for Pmt` where there's a direct and unambiguous representation for T.
* Added missing case for `Pmt::VecCF32` in `PartialEq::eq`

### DIFF
--- a/crates/types/src/pmt.rs
+++ b/crates/types/src/pmt.rs
@@ -228,11 +228,11 @@ impl Pmt {
 #[error("PMt conversion error")]
 pub struct PmtConversionError;
 
-impl TryInto<f64> for &Pmt {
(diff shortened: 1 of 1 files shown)


bastibl:
Awesome, thank you!

PR #162, opened 2024-11-01
Added convenience methods for finding `Block`s in a `Flowgraph`.

metasim:


bastibl:
Great, thank you! Would you be fine with `block_by_name`? That would fit the current naming scheme (https://docs.rs/futuresdr/latest/futuresdr/runtime/stream_io/struct.StreamIo.html#method.input_by_name). I'm not sure, if this is the best though. Are there  best practices in the ecosystem that we could follow?

metasim:
>  Would you be fine with `block_by_name`? T

Absolutely. Updated.

> I'm not sure, if this is the best though. Are there best practices in the ecosystem that we could follow?

I think `_by` is succinct, makes sense, and is consistent with `min_by` and `sort_by`, etc. in `std.

bastibl:
Thanks!

PR #163, opened 2024-11-01
Added ability to query current Source/Sink config via message.

metasim:
The purpose of this PR is to provide the ability to send a message to the `seify::{Source|Sink}` blocks and get a `Config` with the currently active configuration settings. This is useful for remote operation where it may be important to represent the current settings in a user interface or other coordinating software.

### DIFF
--- a/src/blocks/seify/config.rs
+++ b/src/blocks/seify/config.rs
@@ -1,6 +1,7 @@
 use seify::Device;
 use seify::DeviceTrait;
 use seify::Direction;
+use std::collections::HashMap;
 
 use crate::runtime::Error;
 use crate::runtime::Pmt;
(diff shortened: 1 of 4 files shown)


src/blocks/seify/config.rs
```diff
@@ -31,6 +32,27 @@ impl Config {
         Pmt::Any(Box::new(self.clone()))
     }
 
+    /// Generate a [`Pmt`] that can be serialized
+    pub fn to_serializable_pmt(&self) -> Pmt {
```
metasim, line 36: Not a huge fan of this name, but needed to be able to distinguish between the `to_pmt` method whose result is *not* serializable over a remote connection.

metasim: Another option is to replace the `to_pmt` method with this one. The `Config::try_from(Pmt)` function supports both forms.

tests/seify.rs
```diff
@@ -152,13 +159,25 @@ fn config_cmd_map() -> Result<()> {
...
+    match conf {
+        Pmt::VecPmt(v) => match v.as_slice() {
+            [Pmt::MapStrPmt(m), ..] => {
+                assert_eq!(m.get("freq").unwrap(), &Pmt::F64(102e6));
+                assert_eq!(m.get("sample_rate").unwrap(), &Pmt::F64(1e6));
+            }
+            o => panic!("unexpected pmt type {o:?}"),
+        },
+        o => panic!("unexpected pmt type {o:?}"),
+    }
```
metasim, line 181: I coopted an existing test to test this functionality out. Can extract into a separate test if preferred.

bastibl:
`Pmt`s are used to interact with message handlers, since the handlers need to have a unified function signature. If or how they are serialized is a different question. Since most remote APIs are web-based, JSON serialization was natural. But that's independent from encoding a complex type, which should always just interact with PMTs and not w/ JSON strings.

If you want to use a complex type w/ PMTs, you'd come up with some way to store the data in the PMT and implement `From<Config> for Pmt` and `TryFrom<Pmt> for Config`.

bastibl:
Is there a reason why one message handler sends the configs for all channels and the other one expects to get a channel number?

metasim:
> Is there a reason why one message handler sends the configs for all channels and the other one expects to get a channel number?

Oooops. A mistake. They should both require the channel id.

src/blocks/seify/config.rs
metasim: Was the `Box`-ed variant there for performance, convenience, or some other requirement?

> Another option is to replace the `to_pmt` method with this one. The `Config::try_from(Pmt)` function supports both forms.

Was the `Box`-ed variant there for performance, convenience, or some other requirement?

bastibl: IIRC, `Any` requires a `Box`. I wouldn't know of another way. This variant can be used within the same process, i.e., casting it back to `Config`, instead of serialization and deserialization.

metasim: That part makes sense. It was more of a question of whether this is the right approach to supporting both modalities (in-process vs. inter-process). Probably fine for now.

bastibl:
Thanks!

PR #164, opened 2024-11-04
Added ability to query `seify` source capabilities, such as frequency range, sample rates, etc.

metasim:
_NOTE: I need a better understanding of the design goals around sending structures across network channels, i.e. is it `Pmt`-only and the client-side needs to transform into structures, or "serde" over Pmt, or `serde-json`?_

(PR base repository should be the branch for https://github.com/FutureSDR/FutureSDR/pull/163)

WIP; Awaiting: 

* https://github.com/FutureSDR/seify/pull/14
* https://github.com/FutureSDR/FutureSDR/pull/161 
* https://github.com/FutureSDR/FutureSDR/pull/163

## Issues:

* `Range` exists in `seify`, and is only available with the `seify` feature.
* The `Capabilities` type herein needs to be available to both `futuresdr` and `futuresdr-remote`. `futuresdr-types` would be the appropriate place for shared types, but to get the `Range` type would require adding a _feature_ to `futuresdr-types`, pulling in code that shouldn't be a part of a remote client.
* Possible options (none great):
  - Mirror `Range` in a serde-compatible fashion in `futuresdr-types`.
  - Use some other type than `seify::Range` (std `Range` is inadequate)
  - Have `seify` depend on `futuresdr-types`
  - Create a `Pmt` variant for ranges.
  - Leave it to clients to deserialize from [the nested `Pmt` structures](https://github.com/FutureSDR/FutureSDR/pull/164/files#diff-b1b064d2b6f6994a83568d1e539cb119fc034f70d9a3d6192a49ab5ec258ca0cR141-R197).
  - Consider a completely different approach to this capability.
  - Abort adding this to FutureSDR, and implement own custom source.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,6 +74,10 @@ required-features = ["tpb_scheduler"]
 name = "seify"
 required-features = ["seify_dummy"]
 
+[[test]]
+name = "capabilities"
+required-features = ["seify", "soapy"]
+
 [dependencies]
 anyhow = "1.0"
 async-trait = "0.1"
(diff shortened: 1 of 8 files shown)


metasim:
Will rewrite and resubmit after #163 is wrapped up.

PR #168, opened 2024-11-07
Added ability to get `Tag`s generated by test target from `Mocker`.

metasim:


src/runtime/mocker.rs
```diff
@@ -192,10 +237,11 @@ impl<T: Debug + Send + 'static> BufferWriterHost for MockWriter<T> {
         self
     }
 
-    fn produce(&mut self, amount: usize, _tags: Vec<ItemTag>) {
+    fn produce(&mut self, amount: usize, tags: Vec<ItemTag>) {
         unsafe {
             self.data.set_len(self.data.len() + amount);
         }
+        self.tags.extend(tags);
```
bastibl, line 244: I didn't test it, but I guess the index of the tags should be offset by `self.data.len()` before it is increased.

metasim: @bastibl I think this fixes it: [5f1988b](https://github.com/FutureSDR/FutureSDR/pull/168/commits/5f1988beed283aeee1cfdbcd7fce3507436110c9)

bastibl:
It looks like, with these changes `run()` must be called exactly once. How about consuming self and returning the output buffers /w tags?

metasim:
> It looks like, with these changes `run()` must be called exactly once. How about consuming self and returning the output buffers /w tags?

Are you concluding this because of the calls to `init`/`deinit` (which I added because I was testing a source-like block that used `init`).  If so, was adding them like this the right decision?

If that's the right approach, I'd like to make sure I understand what you're asking for before doing it. Here's how I'm interpreting your suggestion:

1. `run` and `run_async` consume `self`
2. Both return something like `(Vec<Vec<T>>, Vec<Vec<ItemTag>>)` (or a named `struct` for readability.
3. To support (2), introduce the type parameter `T: Debug + Send + 'static` to `run` and `run_async`.

bastibl:
I don't have a strong opinion. I just wondered what use-case you had in mind, since this works:
```rust
    mock.init_output::<f32>(0, input.len() * 2);
    mock.input(0, input.clone());
    mock.input_with_tags(0, input, tags.clone());
    mock.run();
```
but this doesn't:
```rust
    mock.init_output::<f32>(0, input.len() * 2);
    mock.input(0, input.clone());
    mock.run();
    mock.input_with_tags(0, input, tags.clone());
    mock.run();
```
In combination with the added `init`, `deinit` calls, I assumed that you envisioned that the mocker is constructed, I/O buffers are set up, and `run()` must be called exactly once. This was not clear to me and the API does not enforce this, which is why I was suggesting that it should be a consuming call. Using the `MockOutput` sounds like a good idea to me.

The other option would be that `run` can be called multiple times (like in the second code example), but then `init` and `deinit` are somewhat unexpected for me and the `input_with_tags` method needs to be fixed (it has to consider the index, i.e., already consumed items when calculating the tag offset). Also the output buffer can only be queried once at the moment, since it `take()`s the data.

So overall, I believe that the first option better fits the current implementation.

metasim:
Let me run one more idea past you before implementing option 1 above. I'm particularly reluctant to get rid of the ability to run the event loop multiple times. (I have a test where I call `run` more than once to ensure my continuation logic is correct.)

* Store a `is_init` flag to track if the wrapped Block has had `init` called on it.
* Call `init` in `Mocker::run` if `is_init` is `false`, then change it to `true`
* Call `deinit` in `Drop`
* Merge `output` and `output_tags` (I think), and change name to `take_output`. 
  - Another option (a variant of above) is to have `run` return a `MockResult` containing the output buffers, but *not* consume `self`.
* Get your assistance with fixing the offsets in `input_with_tags`... I still haven't fully grasped the semantics of those fields. :/ 

What makes this hard is not knowing all the kinds of test cases one might need to exercise with it, or even the most common 80%. That said, the goal is to do black-box testing on a `Block`, which I think comes down to:

1. Specifying zero or more input buffers and messages
2. Running the event loop one or more times.
3. Inspecting the output buffers after each event loop run.

Seems like we're close, except for message handling?

bastibl:
It's not fully tested, but I think it goes in the right direction :-) So let's merge this. Let me know, if you encounter any problems with your project. I'm not using the mocker at the moment.

metasim:
Thanks very much for your help with this.

PR #170, opened 2024-11-12
Improved error reporting when a port is mis-specified on a `Block`.

metasim:
Update to provide more human-friendly error messages on invalid port specifications:

| Context | Before | After |
|--------|--------|--------|
| Stream | `Block 2 does not have stream port (Name("inner"))` | `Block 'Source' does not have stream port 'inner'` |
| Message | `Block Some(5) does not have message port (Name("out2"))`  | `Block 'MessageSource' does not have message port 'out2'` |


bastibl:
Thank you!

PR #172, opened 2024-11-13
Ensure `CurrentInput` has been initialized on call to `StreamIO::tags`

### LINKED ISSUE #171: Implicit assumption in `StreamIO` that buffer is fetched before tags.
Assuming upstream `Block` generates `Tag`s, this fragment at the top of a `Kernel::work` implementation prints nothing:
```rust
sio.input(0).tags().iter().for_each(|t| {
    dbg!(t);
});
let input = sio.input(0).slice::<f32>();
```

While this prints tags as expected:
```rust
let input = sio.input(0).slice::<f32>();
sio.input(0).tags().iter().for_each(|t| {
    dbg!(t);
});
```

This is either a documentation issue or a minor bug, but mainly surprising until one figures it out. Logging it for awareness.

metasim:
Closes #171 

I manually tested this with some existing code. Until #168 is finished, I think it's hard to test this in a regression test.

### DIFF
--- a/src/runtime/stream_io.rs
+++ b/src/runtime/stream_io.rs
@@ -97,14 +97,8 @@ impl StreamInput {
         self.tags.iter_mut().for_each(|x| x.index -= amount);
     }
 
-    /// Get buffer content as slice
-    pub fn slice<T>(&mut self) -> &'static [T] {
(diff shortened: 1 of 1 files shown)


bastibl:
Thanks for fixing this!

PR #173, opened 2024-11-13
Changed `Topology::add_block` and dependents to return a `Result<usize>`

### LINKED ISSUE #169: User-provided value to `BlockMeta::instance_name` is not preserved after insertion into `Flowgraph`.
I was looking to use the `BlockMeta::instance_name` as a means of assigning application-specific unique identifiers to particular stages in the flow graph for deterministic lookup via `Remote`. The following test fails, indicating the user-specified instance name is overwritten by the internals, making the `Block::set_instance_name` method unuseful.

## Test

```rust
use futuresdr::anyhow::Result;
use futuresdr::blocks::NullSink;
use futuresdr::blocks::NullSource;
use futuresdr::runtime::Flowgraph;
use futuresdr::runtime::Runtime;
use futuresdr_macros::connect;

#[test]
fn flowgraph_instance_name() -> Result<()> {
    let rt = Runtime::new();
    let name = "my_special_name";

    let mut source = NullSource::<f32>::new();
    let sink = NullSink::<f32>::new();
    source.set_instance_name(name);
    connect!(fg, source > sink);
    let (th, mut fg) = rt.start_sync(fg);

    let desc = rt.block_on(async move { fg.description().await })?;
    assert_eq!(desc.blocks.first().unwrap().instance_name, name);
    Ok(())
}
```

## Culprit

Line 124 below overwrites the user-assigned value.

https://github.com/FutureSDR/FutureSDR/blob/34bf04f7447980f0b56ad8d407499dd4397c7010/src/runtime/topology.rs#L120-L126

## Possible solution

* Only set instance_name in `Topology::add_block` when it is provided as `None`.
* If `Block::instance_name()` is `Some(..)`, then check `Topology::blocks` for instance name uniqueness.
  - `Topology::add_block` does not return `Result` so the only error handling option is to `panic`, or break the API, or issue a log warning.
  - Implication of the uniqueness constraint not being held are `Remote` queries for a `Block` by name make provide unexpected results.

metasim:
This enables propagation of an error when `insatance_name` is specified, but is not unique in `Flowgraph`.

Closes #169


metasim:
@bastibl This is a pretty extensive change. Let me know if you have second thoughts.

crates/macros/src/lib.rs
```diff
@@ -165,17 +165,17 @@ pub fn connect(attr: proc_macro::TokenStream) -> proc_macro::TokenStream {
         use futuresdr::runtime::Block;
         use futuresdr::runtime::Flowgraph;
 
-        struct Foo;
+        struct FgOp;
```
metasim, line 168: I hope you're ok with this change. When I first saw `Foo` in an expanded macro I thought maybe something inadvertant leaked in. :D

src/runtime/topology.rs
```diff
@@ -118,11 +119,18 @@ impl Topology {
     }
 
     /// Adds a [Block] to the [Topology] returning the `id` of the [Block] in the [Topology].
-    pub fn add_block(&mut self, mut block: Block) -> usize {
-        let block_name = block.type_name();
-        let block_id = self.blocks.vacant_key();
-        block.set_instance_name(format!("{}-{}", block_name, block_id));
-        self.blocks.insert(Some(block))
+    pub fn add_block(&mut self, mut block: Block) -> Result<usize> {
```
bastibl, line 122: This should be a normal `Result<usize, crate::runtime::Error>`, not an `anyhow::Result`.

src/runtime/topology.rs
```diff
@@ -118,11 +119,18 @@ impl Topology {
...
+        if let Some(name) = block.instance_name() {
+            if self.block_id(name).is_some() {
+                Err(Error::DuplicateBlockName(name.to_string()))?;
```
bastibl, line 125: Didn't know that this works. Can we do `return Err(...)`?

bastibl (commented):
Thank you very much! I just have two small comments otherwise I'd say it's good to go. Are you also fine with this change?

src/runtime/topology.rs
metasim: Good catch... I've not been paying attention to the "error" contexts as I should.

src/runtime/topology.rs
metasim: It works, but that combination of using `anyhow::Result` and `?` results in an implicit conversion. Another reason to use `return` instead.

bastibl:
Thanks!

metasim:
For future cleanup: https://github.com/FutureSDR/FutureSDR/issues/175

PR #174, opened 2024-11-14
Added more supported `Pmt` variants to `WebsocketPmtSink`

metasim:
Added support for:

* `Pmt::Blob`
* `Pmt::VecF32` 
* `Pmt::VecU64` 
* `Pmt::String`

I think this covers all the non-scalar variants. Can add the scalar ones if a decision is made on how they should be encoded.


bastibl:
Thanks!

PR #176, opened 2024-11-15
Added convenience to `TagAny` for testing against erased type.

metasim:
Helps with filtering tags relevant for a particular `Kernel`.


bastibl:
Thanks!

PR #179, opened 2024-11-21
Added support for getting the PmtKind from Pmt.

metasim:
Also tweaked `check.sh` and `lttng` features to support other development environments. Can split out in a separate PR if that's annoying.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,7 +1,7 @@
 use rustc_version::version_meta;
 use rustc_version::Channel;
 
-#[cfg(feature = "lttng")]
+#[cfg(all(feature = "lttng", target_os = "linux"))]
 fn gen_lttng_tracepoints() {
     use lttng_ust_generate::CIntegerType;
     use lttng_ust_generate::CTFType;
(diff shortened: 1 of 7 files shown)


build.rs
```diff
@@ -1,7 +1,7 @@
 use rustc_version::version_meta;
 use rustc_version::Channel;
 
-#[cfg(feature = "lttng")]
+#[cfg(all(feature = "lttng", target_os = "linux"))]
```
metasim, line 4: Couldn't figure out how to do this with `Cargo.toml` alone. But kinda feels like there should be a way.

check.sh
```diff
@@ -8,53 +8,55 @@ SCRIPTPATH=`dirname $SCRIPT`
 export RUSTFLAGS='--cfg=web_sys_unstable_apis'
 cd ${SCRIPTPATH} && find . -name "Cargo.lock" -delete
 
+CARGO_FMT="cargo +nightly fmt"
```
metasim, line 11: `+nightly` is needed to support `imports_granularity = "Item"`

bastibl:
Thanks. I subscribed to the wasm bindgen issue for the lint problem. Maybe, we can remove it soon. https://github.com/rustwasm/wasm-bindgen/issues/4283

PR #180, opened 2024-11-21
Added `Pmt::Range`.

metasim:
Good for sending messages reporting upper and lower bounds of values without resorting to serializing to `Pmt::MapStrPmt`.

### DIFF
--- a/crates/types/src/pmt.rs
+++ b/crates/types/src/pmt.rs
@@ -5,6 +5,7 @@ use serde::Serialize;
 use std::any::Any;
 use std::collections::HashMap;
 use std::fmt;
+use std::ops::Range;
 use thiserror::Error;
 
 /// PMT Any trait
@@ -79,6 +80,8 @@ pub enum Pmt {
     F32(f32),
     /// F64, 64-bit float
     F64(f64),
+    /// Range of 64-bit floats
+    Range(Range<f64>),
     /// Vector of 32-bit complex floats.
     VecCF32(Vec<Complex32>),
     /// Vector of 32-bit floats.
@@ -117,6 +120,7 @@ impl fmt::Display for Pmt {
             Pmt::U64(v) => write!(f, "{}", v),
             Pmt::F32(v) => write!(f, "{}", v),
             Pmt::F64(v) => write!(f, "{}", v),
+            Pmt::Range(v) => write!(f, "{:?}", v),
             Pmt::VecCF32(v) => write!(f, "{:?}", v),
             Pmt::VecF32(v) => write!(f, "{:?}", v),
             Pmt::VecU64(v) => write!(f, "{:?}", v),
@@ -143,6 +147,7 @@ impl PartialEq for Pmt {
             (Pmt::U64(x), Pmt::U64(y)) => x == y,
             (Pmt::F32(x), Pmt::F32(y)) => x == y,
             (Pmt::F64(x), Pmt::F64(y)) => x == y,
+            (Pmt::Range(x), Pmt::Range(y)) => x == y,
             (Pmt::VecF32(x), Pmt::VecF32(y)) => x == y,
             (Pmt::VecU64(x), Pmt::VecU64(y)) => x == y,
             (Pmt::VecCF32(x), Pmt::VecCF32(y)) => x == y,
@@ -350,6 +355,28 @@ impl TryFrom<Pmt> for Vec<Complex32> {
     }
 }
 
+impl TryFrom<Pmt> for Range<f64> {
+    type Error = PmtConversionError;
+
+    fn try_from(value: Pmt) -> Result<Range<f64>, Self::Error> {
+        match value {
+            Pmt::Range(r) => Ok(r),
+            _ => Err(PmtConversionError),
+        }
+    }
+}
+
+impl TryFrom<&Pmt> for Range<f64> {
+    type Error = PmtConversionError;
+
+    fn try_from(value: &Pmt) -> Result<Range<f64>, Self::Error> {
+        match value {
+            Pmt::Range(r) => Ok(r.clone()),
+            _ => Err(PmtConversionError),
+        }
+    }
+}
+
 impl TryFrom<Pmt> for Vec<u64> {
     type Error = PmtConversionError;
 
@@ -421,6 +448,12 @@ impl From<Vec<u64>> for Pmt {
     }
 }
 
+impl From<Range<f64>> for Pmt {
+    fn from(r: Range<f64>) -> Self {
+        Pmt::Range(r)
+    }
+}
+
 impl From<Vec<Complex32>> for Pmt {
     fn from(v: Vec<Complex32>) -> Self {
         Pmt::VecCF32(v)
@@ -455,6 +488,8 @@ pub enum PmtKind {
     F32,
     /// F64
     F64,
+    /// Range
+    Range,
     /// VecCF32
     VecCF32,
     /// VecF32
@@ -487,6 +522,7 @@ impl fmt::Display for PmtKind {
             PmtKind::U64 => write!(f, "U64"),
             PmtKind::F32 => write!(f, "F32"),
             PmtKind::F64 => write!(f, "F64"),
+            PmtKind::Range => write!(f, "Range"),
             PmtKind::VecCF32 => write!(f, "VecCF32"),
             PmtKind::VecF32 => write!(f, "VecF32"),
             PmtKind::VecU64 => write!(f, "VecU64"),
@@ -515,6 +551,7 @@ impl std::str::FromStr for PmtKind {
             "U64" => return Ok(PmtKind::U64),
             "F32" => return Ok(PmtKind::F32),
             "F64" => return Ok(PmtKind::F64),
+            "Range" => return Ok(PmtKind::Range),
             "VecF32" => return Ok(PmtKind::VecF32),
             "VecU64" => return Ok(PmtKind::VecU64),
             "Blob" => return Ok(PmtKind::Blob),
@@ -675,6 +712,11 @@ mod test {
         let p = Pmt::from(e);
         assert_eq!(p, Pmt::Bool(e));
         assert_eq!((&p).try_into(), Ok(e));
+
+        let e = 34f64..42f64;
+        let p = Pmt::from(e.clone());
+        assert_eq!(p, Pmt::Range(e.clone()));
+        assert_eq!(p.try_into(), Ok(e));
 
         let e = vec![1.0, 2.0, 3.0];
         let p = Pmt::from(e.clone());


bastibl:
I am not sure what to do here. I think, we need a better idea about the role of Pmts. At least until now, I saw them more or as a data store. Range, in turn, adds some semantics. Instead of a Range, should we maybe just add a tuple type? Or expect the user to encode a Range as a Pmt with a Vec<f64> that has two items? In any case, we could implement `TryFrom<Pmt> for Range<f64>` etc.

If we really want to add a Range, maybe it should be custom type that also allows step sizes? I think that's pretty common for parameter ranges of devices, for examples. See: https://docs.rs/seify/latest/seify/struct.Range.html

metasim:
> I am not sure what to do here. I think, we need a better idea about the role of Pmts. At least until now, I saw them more or as a data store. Range, in turn, adds some semantics. 

I'm very sympathetic to that. To me it feels to me like `Pmt` serves multiple roles: status response (`Ok`, `Finished`, `InvalidValue`); and data (the rest of the variants). `Pmt` also serves these roles under two contexts: inter-block communications (where low overhead is important) and remote/network communications (where serialization of some sort is needed). More on that below.

To determine what we do with this PR, I'll outline my use case (which is remote/network communications of low-throughput data) and we can see if that helps flesh out a better approach. 


## SDR Capabilities Use Case

I need the ability to query the seify driver for basic "capabilities" like frequency range, sample rates, agc support, etc. I briefly [submitted a PR](https://github.com/FutureSDR/FutureSDR/pull/164) on this, but yanked it because I wasnt' happy with the design. The main issue was serialization of seify's `Range`/`RangeItem` types. 

`Range` lives in the `seify` crate, and `Pmt` lives in the `futuresdr-types` crate, which are (correctly) orthogonal namespaces. Due to the orphan rule, converting from `Pmt` and seify `Range` can only be implemented in something that depends on both. In an attempt to solve that problem I ended up with this monstrosity:


```rust
/// Newtype to assist in converting between a seify [`Range`] and [`Pmt`].
struct Conv<'a, T>(&'a T);

impl<'a> From<Conv<'a, Range>> for Pmt {
    fn from(range: Conv<'a, Range>) -> Self {
        Pmt::VecPmt(
            range
                .0
                .items
                .iter()
                .map(|x| match x {
                    RangeItem::Interval(min, max) => Pmt::MapStrPmt(HashMap::from([
                        ("min".to_owned(), Pmt::F64(*min)),
                        ("max".to_owned(), Pmt::F64(*max)),
                    ])),
                    RangeItem::Value(v) => Pmt::F64(*v),
                    RangeItem::Step(min, max, step) => Pmt::MapStrPmt(HashMap::from([
                        ("min".to_owned(), Pmt::F64(*min)),
                        ("max".to_owned(), Pmt::F64(*max)),
                        ("step".to_owned(), Pmt::F64(*step)),
                    ])),
                })
                .collect(),
        )
    }
}

impl<'a> TryFrom<Conv<'a, Pmt>> for Range {
    type Error = Error;

    fn try_from(value: Conv<'a, Pmt>) -> Result<Self> {
        match value.0 {
            Pmt::VecPmt(v) => {
                let items = v
                    .iter()
                    .map(|x| match x {
                        Pmt::MapStrPmt(m) => {
                            let min: f64 =
                                m.get("min").context("missing min")?.to_owned().try_into()?;
                            let max = m.get("max").context("missing max")?.to_owned().try_into()?;
                            let step = m.get("step");
                            if let Some(step) = step {
                                Ok(RangeItem::Step(
                                    min,
                                    max,
                                    step.to_owned().try_into().context("step not f64")?,
                                ))
                            } else {
                                Ok(RangeItem::Interval(min, max))
                            }
                        }
                        Pmt::F64(v) => Ok(RangeItem::Value(*v)),
                        _ => Err(Error::msg("unexpected pmt type")),
                    })
                    .collect::<Result<Vec<RangeItem>>>()?;
                Ok(Range { items })
            }
            o => Err(Error::msg(format!("unexpected Pmt value: {:?}", o))),
        }
    }
}

...
//  convert from `Range` to `Pmt`.
let range = Range { items: vec![RangeItem::Interval(100.0, 200.0)] };
let pmt: Pmt =  Conv(range).into();
let range: Range =  pmt.try_into()?;
```

It has a smell of lots of accidental complexity, hard to maintain, and doesn't scale. Yuck.

I then wondered if I should implement a `serde_pmt` crate, since most relevant types are `Serializable` already, but that's too much work, and feels like duplicating what `serde_json` already does. 

Then I considered if we should have a `Pmt::Json(serde_json::Value)` variant. But since `serde_json::Value` will need to be further serialized to a `String`, might as well just use `Pmt::String` although I'm still on the fence about that (having a `Pmt::Json` variant at least tells the remote endpoint the format of the payload).

Finally, I went with what I thought would be the minimal API change which was adding `Pmt::Range(Range<f64>)`. I almost made it `Pmt::Range(Range<f64>, Option<f64>)` to cover the step case. And that's how I got to this PR.

## Alternatives

> Instead of a Range, should we maybe just add a tuple type? 

The first, maybe, but it's not bringing much more interpretability to the table than a `Pmt::VecF64`

> Or expect the user to encode a Range as a Pmt with a Vec that has two items? 

The second ends up with the mess above. Neither feel satisfying to me. Plus, what's size and type of the tuple?

> In any case, We could implement TryFrom<Pmt> for Range<f64> etc.

That definitely moves the ball forward. We can incrementally add to/from implementations of things very easily as long as they fall into the namespace of `futuresdr-types`, but application-specific types get ugly like the above. You have to hand-craft serde operations specifically around `Pmt`. 

After writing this up, I think what _I_ really want is just something that's `serde` compatible. Almost everything supports it, is maintainable, and has low ceremony. But I don't know how to get there in the short term.

## Path Forward

For my personal use case, if `Pmt::Range` doesn't feel right I can probably do what I need with `Pmt::String`. I didn't go that route initially because I've tried to approach my use of FutureSDR more generally, filling in gaps as I find them in ways that I hope are useful to others, and using `Pmt::String` feels like a backdoor around capabilities `Pmt` should be able to handle. 

Having some form of `Range<f64>` support gets me further without resorting to `serde_json::to_string` in my `#[message_handler]`s, but suspect I'll run into other gaps in my work ahead. 

Therefore, for this PR, I see the following options based on the above (please advise):

* Keep `Pmt::Range` for now, and consider a better solution in another PR.
* Remove `Pmt::Range` and instead implement `From<Range<f64>> for Pmt` and  `TryFrom<Pmt> for Range<f64>`
* Remove `Pmt::Range` and implement `Pmt::F64Tuple(f64, f64)`
* Remove `Pmt::Range` and implement `Pmt::VecF64`
* Remove `Pmt::Range` and implement (possibly behind a `pmt_json` feature) `Pmt::Json(serde_json::Value)`
* Retract this PR altogether and use `Pmt::String` with my application-specific JSON payloads.
* Try to address the larger question of "what is the role of `Pmt`" and implement a longer term solution (see next section).

I'm amenalble to all the above. I thihk it comes down to what your goals and vision for the project are.

---

## Future Food for Thought

If we want to discuss the "what is the role of `Pmt`" question, here's an outline of my current thoughts.

I've assumed the design of `Pmt` is influenced by GnuRadio, which I only have passing exposure to (I'm new to the SDR field), so my assumptions about the role of `Pmt` are likely naive. But from my outside perspective I see the "`Pmt` problem space" like this:

* local
    - inter-block message passing
        - push small and large data packets between blocks with minimal overhead (e.g. no serialization)
        - return status response (`Ok`, `InvalidValue`, etc.)
* remote
    - network query and RPC-like operations
        - remote apply setting with ack/nack (i.e. `call` with `Ok` & `InvalidValue`)
        - remote request/response (i.e. `callback` with serialized data structures)
        - push structured data to remote via `WebsocketPmtSink`

How that maps to the `Pmt` goals, I'm not sure, but to me feels like `Pmt` is optimized for the first section (local), and for remote you really want JSON via `serde` (IMHO). If that's correct, the consequence is that `#[message_handler]`s (or the code that calls them) have to accommodate the local/remote contexts somehow, or the `#[message_handler]`s are bifurcated into two types/contexts. There's probably other more elegant solutions I'm not thinking of via traits, like `axum` does with `IntoResponse`.

One way of accommodating that might be to revisit the data modeling to better reflect the contexts it operates in.

```rust
enum Pmt2 {
    Status(PmtStatus),
    Payload(PmtPayload),
}

enum PmtStatus {
    Ok,
    InvalidValue,
    Finished,
}

enum PmtPayload {
    Native(PmtNative),
    Serialized(Box<serde_json::Value>),
    Any(Box<dyn PmtAny>),
}

enum PmtNative {
    Null,
    String(String),
    Bool(bool),
    Usize(usize),
    Isize(isize),
    U32(u32),
    U64(u64),
    F32(f32),
    F64(f64),
    VecCF32(Vec<Complex32>),
    VecF32(Vec<f32>),
    VecU64(Vec<u64>),
    Blob(Vec<u8>),
    VecPmt(Vec<Pmt>),
    MapStrPmt(HashMap<String, Pmt>),
}
```

Regardless, of whether any of this is on the mark, it's getting into larger architectural changes which may or may not be within the current project goals, and I don't know the library architecture well enough yet to know the implications.

bastibl:
Thank you for the detailed reply. You're right that PMTs server multiple purposes, but I still hope to align them without too much headaches.

Regarding the seify range. This is only an issue since this application-specific type comes from another crate (i.e., seify). If this would be a type from your crate, you wouldn't need the new type pattern. Also the complexity of the code that you pasted is really mainly the conversion between the PMT and the seify range. I don't think that the wrapper type adds too much. So I would say that this is perfectly fine, in general. In this specific case, since seify is already a dependency, I'm happy to add a TryFrom implementation with a feature flag (we just need to add a feature to the types crate and propagate it from the main crate).

Regarding JSON serialization: I don't think that we should add a variant for JSON serialized data. Then it would no longer be obvious how to encode data. PMTs and JSON should be converted between each other, but they should not contain each other. I think to encode everything that JSON can encode, PMTs would need a few more vector types (and I'm happy to add them). The idea is that the web API converts between JSON and PMTs (with serde), and the applications convert between PMTs and application-specific types (with TryFrom).

So to proceed with this PR, I think we should add a PMT <-> seify range conversion to futuresdr-types with a seify feature flag. Would this be fine for you?

metasim:
> Regarding JSON serialization: I don't think that we should add a variant for JSON serialized data. Then it would no longer be obvious how to encode data. PMTs and JSON should be converted between each other, but they should not contain each other. I think to encode everything that JSON can encode, PMTs would need a few more vector types (and I'm happy to add them). The idea is that the web API converts between JSON and PMTs (with serde), and the applications convert between PMTs and application-specific types (with TryFrom).

Thank you. This is a very helpful clarification on where the abstraction boundary sits. 

> So to proceed with this PR, I think we should add a PMT <-> seify range conversion to futuresdr-types with a seify feature flag. Would this be fine for you?

That sounds great!. Do you want me to take a stab at it? If so, do you have in mind a feature flag name? If not, I'm not in a huge hurry: We have a holiday week starting in the USA, and I'm on a deadline, so I'm not going to be able to get back to this for a week or so.

metasim:
Superseded by #182

PR #182, opened 2024-12-03
Added `seify` feature to enable PMT conversions for Seify types.

metasim:
In addition to feature flag, implements `Pmt` conversions for `seify::Range`.

See conversation on [this PR](https://github.com/FutureSDR/FutureSDR/pull/180) for context.

Supersedes #180


.github/workflows/ci.yml
```diff
@@ -392,6 +392,7 @@ jobs:
       - run: sudo apt-get -y install libsoapysdr-dev
       - run: cargo test --all-targets --workspace --features=aaronia_http,rtlsdr,zeromq,audio,flow_scheduler,tpb_scheduler,seify_dummy,soapy,lttng,zynq,wgpu
       - run: cargo test --all-targets --manifest-path=crates/futuredsp/Cargo.toml
+      - run: cargo test --all-targets --all-features --manifest-path=crates/types/Cargo.toml
```
metasim, line 395: Not sure I understand why sub-crate tests don't get run automatically, but this seems to be required to get test in `futuresdr-types` to run?

Cargo.toml
```diff
@@ -29,7 +29,7 @@ flow_scheduler = []
 lttng = ["dep:lttng-ust", "dep:lttng-ust-generate"]
 rtlsdr = ["seify/rtlsdr"]
 hackrf = ["seify/hackrfone"]
-seify = ["dep:seify"]
+seify = ["dep:seify", "futuresdr-types/seify"]
```
metasim, line 32: Assuming if you have `seify` enabled in `futuresdr`, you'd also want the `Pmt` conversion support.

crates/types/Cargo.toml
```diff
@@ -10,9 +10,15 @@ description = "Polymorphic Types for an Experimental Async SDR Runtime for Heter
...
 
+[features]
+default = []
+# Provides support for converting seify types into Pmt variants.
+seify = ["dep:seify"]
+
 [dependencies]
 dyn-clone = "1.0"
 num-complex = { version = "0.4", features = ["serde"] }
+seify = { version = "0.16", default-features = false, optional = true }
```
metasim, line 21: Is there a better way of keeping this seify version in sync with the other crates? I don't see the use of `xyz.workspace = true` in any other `Cargo.toml` files, so assumed there's a reason for that.

metasim:
Build failure is unrelated; guessing a new lint?

image

.github/workflows/ci.yml
bastibl: I never questioned it :-) but I just had a look and there seems to be a workspace flag for tests. However, since only few crates are actually part of the same workspace, I'd say we just do it like this for all crates.

crates/types/Cargo.toml
bastibl: Tbh, I didn't know that this is possible. But it seems that workspace dependencies cannot be optional. So it might not be applicable, but I'm not sure.

bastibl:
Thanks!

crates/types/Cargo.toml
metasim: FWIW, you can tweak the "inheritance" with specific overrides like this:

```toml
seify = { workspace = true, optional = true }
```

I personally find TOML very weird...

PR #185, opened 2024-12-05
Tweak to message handler error propagation to ensure cause is logged.

metasim:


src/runtime/flowgraph.rs
```diff
@@ -157,7 +157,7 @@ impl FlowgraphHandle {
             })
             .await
             .or(Err(Error::InvalidBlock(block_id)))?;
-        rx.await.or(Err(Error::HandlerError))?
+        rx.await?
```
metasim, line 160: Now covered by `Error::ChannelCanceled`

bastibl:
Awesome, thank you!

PR #187, opened 2024-12-13
Adds `take<T: PmtAny>(self: Box<Self>) -> Option<Box<T>>`

### LINKED ISSUE #186: Ability to "take" value out of `PmtAny`
It would be a nice convenience to be able to extract the value inside a `Pmt::Any` to enable a `Block` to take full ownership of the message contents without a post-downcast `clone`.

## Use Case

```rust
let p: Pmt = ...;
let my_foo: Box<Foo> = match p {
    Pmt::Any(payload) => payload.take::<Foo>(),
    _ => panic!(),
};
```

## First Attempt

We currently have this:

https://github.com/FutureSDR/FutureSDR/blob/9d264b27e01970a13ba4f9e24cdc5259227999e0/crates/types/src/pmt.rs#L36-L45

We could extend it with a `take` method _if_ there were a way to get a `Box<dyn Any>` (not `Box<dyn PmtAny>`) via the `PmtAny` trait.

```rust
impl dyn PmtAny {
    ...
    pub fn take<T: PmtAny>(self) -> Option<Box<T>> {
        // if `fn to_any_boxed(self) -> Box<dyn Any>` existed on `PmtAny`
        self.to_any_boxed().downcast().ok()
    }
}
```

I assume this is possible, but I couldn't figure it out (without digging into `unsafe` methods).

metasim:
Closes #186

### DIFF
--- a/crates/types/src/pmt.rs
+++ b/crates/types/src/pmt.rs
@@ -15,6 +15,8 @@ pub trait PmtAny: Any + DynClone + Send + Sync + 'static {
     fn as_any(&self) -> &dyn Any;
     /// Cast to mutable [`Any`](std::any::Any)
     fn as_any_mut(&mut self) -> &mut dyn Any;
+    /// Consume `self`, converting to a `Box<dyn Any>`
+    fn to_any(self: Box<Self>) -> Box<dyn Any>;
 }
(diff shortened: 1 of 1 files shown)


bastibl:
Great, thanks!

PR #188, opened 2024-12-16
Fix multiple initialization prompts on repeated 'start' clicks in examples/cw

randsoy:
Fix multiple initialization prompts on repeated 'start' clicks in examples/cw

### DIFF
--- a/crates/prophecy/src/waterfall.rs
+++ b/crates/prophecy/src/waterfall.rs
@@ -206,7 +206,6 @@ fn render(
             if let Some(bytes) = data.borrow_mut().take() {
                 // assert_eq!(bytes.len(), fft_size_val * 4);
                 fft_size_val = bytes.len() / 4;
-                // 
                 if fft_size_val != last_fft_size {
                     initialize_texture(gl, fft_size_val);
                 }
--- a/crates/prophecy/src/waterfall.rs
+++ b/crates/prophecy/src/waterfall.rs
@@ -37,7 +37,6 @@ const SHADER_HEIGHT: usize = 256;
 pub fn Waterfall(
     #[prop(into)] min: MaybeSignal<f32>,
     #[prop(into)] max: MaybeSignal<f32>,
-    #[prop(into)] fft_size: MaybeSignal<usize>,
     #[prop(optional)] mode: WaterfallMode,
 ) -> impl IntoView {
     let data = match mode {
@@ -117,6 +116,7 @@ pub fn Waterfall(
                     gl_FragColor = vec4(color_map(clamp(power, 0.0, 1.0)), 1.0);
                 }
             ";
+
             let frag_shader = gl.create_shader(GL::FRAGMENT_SHADER).unwrap();
             gl.shader_source(&frag_shader, frag_code);
             gl.compile_shader(&frag_shader);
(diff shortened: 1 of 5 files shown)


examples/app
bastibl, line 1: I assume this was added by mistake?

examples/spectrum/src/wasm/frontend.rs
```diff
@@ -84,10 +86,10 @@ pub fn Spectrum(fg_handle: FlowgraphHandle) -> impl IntoView {
 
     view! {
         <div class="text-white">
-            <button class="bg-slate-600 hover:bg-slate-800 rounded p-2 m-4" on:click=ctrl_click>Show/Hide Controlls</button>
+            <button class="p-2 m-4 rounded bg-slate-600 hover:bg-slate-800" on:click=ctrl_click>Show/Hide Controlls</button>
```
bastibl, line 89: I assume you have a plugin that does that reordering for you. Do you have a recommendation?

crates/prophecy/src/waterfall.rs
```diff
@@ -37,6 +37,7 @@ const SHADER_HEIGHT: usize = 256;
 pub fn Waterfall(
     #[prop(into)] min: MaybeSignal<f32>,
     #[prop(into)] max: MaybeSignal<f32>,
+    #[prop(into)] fft_size: MaybeSignal<usize>,
```
bastibl, line 40: Did you try using the `fft_size` parameter as a signal. It's been a while since I used Leptos, and I wonder whether the component function is actually rerun when the signal changes. IIRC, it does not.

bastibl (commented):
Thank you very much for the PR! I mainly wonder about the use of `fft_size` as a signal. 

Maybe you could also have a look into the CI formatting issue.

examples/spectrum/src/wasm/frontend.rs
randsoy: lepotsfmt:`cargo install leptosfmt`  
rust-analyzer.toml:`[rustfmt]
overrideCommand = ["leptosfmt", "--stdin", "--rustfmt"]`

crates/prophecy/src/waterfall.rs
randsoy: You can refer to min and max, as your re-rendering is using requestAnimationFrame. At least I can modify it through the component parameters, otherwise, it will cause an exception.

examples/app
randsoy: sorry

crates/prophecy/src/waterfall.rs
bastibl: `min` and `max` just change some values that can be modified during runtime, i.e., they can be a Leptos signal. For the FFT, we initialize a static buffer with the corresponding size. If you increase the FFT size later, you'll get a buffer overflow. This means, this parameter cannot be changed during runtime, i.e., it should be a prop but not a Leptos signal. I think, we should just change this to a `usize`.

examples/spectrum/src/wasm/frontend.rs
bastibl: Thanks, didn't know that this will also take care of Tailwind classes.

crates/prophecy/src/waterfall.rs
randsoy: I hope fft_size can also be changed at runtime, I'm thinking about how to change it. Check if the transformation rebinds the gl data. Give me some more time.

bastibl:
The current version adds again a submodule in the examples folder. Also, it seems to me that the current version also hard-codes the `fft_size` but changes it to 1024 from 2048. I'm not sure if this is intended.

randsoy:
> The current version adds again a submodule in the examples folder. Also, it seems to me that the current version also hard-codes the `fft_size` but changes it to 1024 from 2048. I'm not sure if this is intended.

1024 is only an initial value. The real fft_size changes according to the length of the data.

bastibl:
Ah, that's a good idea. Did you try changing the FFT size during runtime?? I don't have a test case right now.

randsoy:
> Ah, that's a good idea. Did you try changing the FFT size during runtime?? I don't have a test case right now.

After testing, you can see that the default value of fft_size in the example is 2048, but I have 1024 here, and it will automatically adapt to 2048

bastibl:
Ok, great. Can you please remove the submodule in the examples folder. Then we can merge it.

bastibl:
Thanks!

PR #193, opened 2025-02-17
Support Zenoh publisher sink and subscriber source.

CaseySanchez:
These Zenoh blocks are based upon the existing ZeroMQ blocks.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ vulkan = ["dep:vulkano", "dep:vulkano-shaders"]
 wgpu = ["dep:wgpu"]
 zeromq = ["dep:zmq"]
 zynq = ["dep:xilinx-dma"]
+zenoh = ["dep:zenoh"]
 
 [[bench]]
 name = "flowgraph"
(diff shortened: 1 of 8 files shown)


CaseySanchez:
@bastibl I most certainly can add it to `fsdr-blocks`, but if I may I would first like for you to consider my rationale for implementing these blocks.

The Redhawk SDR framework leverages the Common Object Request Broker Architecture (CORBA) specification from the Object Management Group (OMG) for IPC. It could be said that CORBA was effectively succeeded by the OMG's Data Distribution Service (DDS) specification. Zenoh has proven itself to be a worthy competitor of DDS, managing to become a supported middleware of ROS2. Furthermore, Zenoh has bindings to many different languages, most notably C++, C, Python, and Typescript.

If you still beg to differ on supporting Zenoh directly in the `FutureSDR` project, I will proceed with opening a PR on `fsdr-blocks`. Please let me know what you think concerning this matter.

bastibl:
I got an email about a reply to my comment, but it is not here? Did you delete it?

PR #209, opened 2025-10-19
Make file_sink and file_source use std::path::Path instead of String

shalen-original:
Changes `FileSink` and `FileSource` constructors to accept (and store) a `Path` instead of a `String`, making the intent clearer.

The change should be mostly backwards-compatible, because `Path` implements `AsRef<Path>` for `String` and `str` ([docs](https://doc.rust-lang.org/std/path/struct.Path.html#impl-AsRef%3CPath%3E-for-String)).
Testing should be covered by the already-existing doctests.

Thanks for the amazing work on FutureSDR!
This is my first contribution, I hope it is welcome. If I have to do something different just let me know, no problem :)

### DIFF
--- a/src/blocks/file_sink.rs
+++ b/src/blocks/file_sink.rs
@@ -1,6 +1,8 @@
 use async_fs::File;
 use futures::io::AsyncWriteExt;
 use std::fs::OpenOptions;
+use std::path::Path;
+use std::path::PathBuf;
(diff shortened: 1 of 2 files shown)


bastibl:
Awesome, thank you! I'll have to look into CI to apply recent clippy lints.

PR #211, opened 2025-12-11
update LoRa example, update PFB, improve HackRF TX

011000101101:
specifically improve SF5 and SF6 performance and enable soft-decoding;
exchange PFB implementations with ones derived from liquid-dsp;
add tagged burst processing blocks to enable corruption-free TX of large frames and immediate TX of small ones, including LoRa example;
add min buffer size setter for seify sink builder; 
also clippy


bastibl:
Awesome, thanks!

PR #218, opened 2026-02-16
feat: implement PartialEq for Tag enum

thief-sty:
Replace the previous impl PartialEq for ItemTag with a Derive

(this is just plain better than #214)


bastibl:
Makes sense, thanks :-)

PR #237, opened 2026-03-31
lora: unify interfaces

011000101101:
add convenience wrappers for RX and TX chain creation, introduce enums for arguments with non-trivial types (sync-word, ldro mode, header mode, meshtastic presets).

main additions in lib.rs, utils.rs and meshtastic.rs, plus necessary updates of interfaces throughout all files.


bastibl:
Thanks for the update! Let's see what CI says with the FutureDSP fix.

PR #240, opened 2026-04-21
Increase TX buffer size for Zigbee example

mucaglar:
This PR increases the TX buffer size for tx and trx modes of Zigbee examples considering the maximum possible sample size of a Zigbee packet.


bastibl:
Thanks!

PR #241, opened 2026-04-24
Prevent confusion between ping and null messages in rattlegram

mucaglar:
This PR fixes the issue where ping messages were incorrectly classified as null packets in the rattlegram example. This ensures that control messages are handled separately from empty data packets.


bastibl:
Thank you!

PR #243, opened 2026-05-04
document seify channel mapping with USRP B210 examples

mucaglar:
This PR documents channel selection in Seify builder with USRP B210 examples to clarify hardware-to-software port mapping.


bastibl:
thanks!

PR #246, opened 2026-07-26
seify: update to new API

narottamroyal:
Switches to the new dynamic types (`DynDevice`/`DynRxStreamer`/`DynTxStreamer`), using the `*Control` traits as necessary.

It could also be nice to implement `From<Device<T>> for DynDevice` in Seify so that `Builder::from_device` can take both typed and dynamic devices. Happy to create a PR if you think it's a good idea.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,7 +82,7 @@ num-complex = "0.4"
 num-integer = "0.1"
 once_cell = "1.21"
 rustfft = "6.4"
-seify = { version = "0.18", default-features = false, optional = true }
+seify = { version = "0.20", default-features = false, optional = true }
 slab = "0.4"
 spin = { version = "0.10", optional = true }
 serde = { version = "1.0", features = ["derive"] }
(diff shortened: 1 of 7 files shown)


bastibl:
Hi, I had some changes in the pipeline for FutureSDR and Seify and took this as an occasion to release them. The main difference to the current PR is that one can use the Seify source/sink in a typed and type-erased (DynDevice) fashion, e.g.,

```rust
let dev = Device::<RtlSdr>::from_args("")?;

let source = Builder::from_device(dev)
  .frequency(100e6)
  .sample_rate(2.4e6)
  .gain(20.0)
  .build_source()?;
```
```rust
let source = Builder::new("driver=rtlsdr")?
  .frequency(100e6)
  .sample_rate(2.4e6)
  .gain(20.0)
  .build_source()?;
```

There should also be better conversions available now. Sorry, that this supersedes this PR.

Let me know if you have any issues with the current state.

